PR #191, opened 2014-03-28
[WIP] pymatbridge  package

cekees:
Installing pymatbridge requires knowing where matlab is installed so that a mex file can be compiled.  This initial version of the package works by hardcoding my matlab's location in a build stage. I think the right way to do this is to also add a host-matlab package, but that still leaves the issue of matlab's location being completely arbitrary, most likely user-dependent. Every user would have to create a new spec just to specify the matlab location. The good news (@arokem, @haoxingz) is that I think this package will work on our cygwin stack.


cekees:
#757 includes the pymatbridge  package  and the host-matlab package

PR #366, opened 2014-07-15
mpich: disable fortran support on osx.

bluescarni:
Since Xcode does not have support for fortran, I propose that we do not build mpich with fortran support on OSX. We can then use ``mpi: use mpich`` to satisfy the mpi dependency of packages such as boost, which currently does not build on osx.


ahmadia:
closing in favor of: https://github.com/hashdist/hashstack/pull/615

PR #427, opened 2014-08-21
Add Firefox

certik:
My browser is too old, so it'd be nice to be able to compile the latest Firefox. Here is info how to compile it:

http://www.linuxfromscratch.org/blfs/view/svn/xsoft/firefox.html

### DIFF
--- a/pkgs/firefox.yaml
+++ b/pkgs/firefox.yaml
@@ -1,7 +1,7 @@
 extends: [base_package, libflags]
 
 dependencies:
-  build: [python]
+  build: [python, gtk2]
 
 sources:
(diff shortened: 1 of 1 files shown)


ahmadia:
Hi @certik - Sorry for the radio silence on this.  Have you had a chance to fix the gtk2 dependency?

certik:
I am back on it. Pixman fails (#670), so I can't build `gtk2` for now. Keep this PR open, the Firefox package should be on the right track, I just need to figure out how to build the dependencies.

certik:
Pixman is fixed by (#671).

Now the big problem: #672. That was the original problem that I was hitting in this PR, and unfortunately it still fails with the latest Hashstack, and I don't know how to fix it.

certik:
Another problem: #674.

PR #598, opened 2014-12-31
initial compatiblity with python 3

ohanar:
This is for use with hashdist/hashdist#302, although unlike that there should be nothing holding this one back.


ahmadia:
Merging.

ahmadia:
Thanks @ohanar

PR #602, opened 2015-01-08
Add mpmath 0.19

certik:
This is needed for the git version of sympy and once sympy is released, for the released version as well.


certik:
The failure is just a timeout and it's unrelated.

PR #608, opened 2015-01-27
Do not set PYTHON_INCLUDE_PATH when building vtk

johannesring:


certik:
I think this will fail if you use the hashdist's python, wouldn't it? If so, we need to add this as an optional parameter that you set in your profile.

johannesring:
It is okay since `${PYTHON_DIR}` is added to `CMAKE_PREFIX_PATH`. I tried it now with hashdist's python and it did pick the correct python include directory and library. I also tried with hashdist's python but removed `${PYTHON_DIR}` from `CMAKE_PREFIX_PATH` and then it did not pick the correct paths but used the ones from the system.

certik:
Ah ok, then I think we are fine.

PR #613, opened 2015-02-02
Add test to parmetis

johannesring:


johannesring:
This did not work on OS X:

    dyld: Library not loaded: libparmetis.dylib
      Referenced from: /Users/johannr/.hashdist/bld/parmetis/evup4xf2ezot/bin/ptest
      Reason: image not found
    dyld: Library not loaded: libparmetis.dylib
      Referenced from: /Users/johannr/.hashdist/bld/parmetis/evup4xf2ezot/bin/ptest
      Reason: image not found
    dyld: Library not loaded: libparmetis.dylib
      Referenced from: /Users/johannr/.hashdist/bld/parmetis/evup4xf2ezot/bin/ptest
      Reason: image not found
    dyld: Library not loaded: libparmetis.dylib
      Referenced from: /Users/johannr/.hashdist/bld/parmetis/evup4xf2ezot/bin/ptest
      Reason: image not found
    --------------------------------------------------------------------------
    mpiexec noticed that process rank 0 with PID 7972 on node Johanness-Mac-3.local exited on signal 5 (Trace/BPT trap: 5).

Let me take a closer look.

johannesring:
It looks like the rpath has been set correctly:

    [HASHDIST DEBUG] johannr@Johanness-Mac-3:parmetis-evup4xf2ezot-1$ otool -l ${ARTIFACT}/bin/ptest | grep LC_RPATH -A2
              cmd LC_RPATH
          cmdsize 64
             path /Users/johannr/.hashdist/bld/mpi/qagb63ozbmab/lib (offset 12)
    --
              cmd LC_RPATH
          cmdsize 72
             path /Users/johannr/.hashdist/bld/parmetis/evup4xf2ezot/lib (offset 12)

However, it works if I add `${ARTIFACT}/lib` to `DYLD_FALLBACK_LIBRARY_PATH`:

    [HASHDIST DEBUG] johannr@Johanness-Mac-3:parmetis-evup4xf2ezot-1$ export DYLD_FALLBACK_LIBRARY_PATH=${ARTIFACT}/lib
    [HASHDIST DEBUG] johannr@Johanness-Mac-3:parmetis-evup4xf2ezot-1$ ${MPIEXEC} -np 4 /Users/johannr/.hashdist/bld/parmetis/evup4xf2ezot/bin/ptest Graphs/rotor.graph
    
    Testing ParMETIS_V3_PartKway with ncon: 1, nparts: 8
         Setup: Max:   0.077, Sum:   0.258, Balance:   1.194
       Matching: Max:   0.049, Sum:   0.197, Balance:   1.004
    Contraction: Max:   0.109, Sum:   0.435, Balance:   1.006
       InitPart: Max:   0.020, Sum:   0.078, Balance:   1.001
        Project: Max:   0.002, Sum:   0.006, Balance:   1.349
     Initialize: Max:   0.020, Sum:   0.079, Balance:   1.013
          K-way: Max:   0.047, Sum:   0.181, Balance:   1.031
          Remap: Max:   0.000, Sum:   0.000, Balance:   1.538
          Total: Max:   0.341, Sum:   1.362, Balance:   1.001
    Final   8-way Cut:  14978 	Balance: 1.045 
    ParMETIS_V3_PartKway reported a cut of 14978 [OK:14978]
    ...

Any idea?

certik:
Maybe @ahmadia would know, I only have experience with these kinds of bugs on linux.

ahmadia:
Do a complete check of dependencies beneath parmetis.  I'm guessing one of the libraries it is referencing is failing to load, causing the error you're seeing.

ahmadia:
Ping me again if you're still stuck and I'll try to take a look.

johannesring:
Thanks @ahmadia. The only dependency for `parmetis` is `mpi` and I am using `openmpi` built with HashDist. I still have the same problem and I am not sure what to try, so it would be greatly appreciated if you could take a look.

ahmadia:
The `ptest` binary isn't being built with proper rpath support.  Note:

```
[HASHDIST DEBUG] otool -L ~/.hashdist/bld/parmetis/fqhqgo2cuzxn/bin/ptest
/Users/aron/.hashdist/bld/parmetis/fqhqgo2cuzxn/bin/ptest:
	libparmetis.dylib (compatibility version 0.0.0, current version 0.0.0)
	/usr/lib/libSystem.B.dylib (compatibility version 1.0.0, current version 1197.1.1)
	/Users/aron/.hashdist/bld/mpi/jsrkktwepaqt/lib/libpmpich.10.dylib (compatibility version 11.0.0, current version 11.3.0)
	/Users/aron/.hashdist/bld/mpi/jsrkktwepaqt/lib/libmpich.10.dylib (compatibility version 11.0.0, current version 11.3.0)
	/Users/aron/.hashdist/bld/mpi/jsrkktwepaqt/lib/libopa.1.dylib (compatibility version 2.0.0, current version 2.0.0)
	/Users/aron/.hashdist/bld/mpi/jsrkktwepaqt/lib/libmpl.1.dylib (compatibility version 2.0.0, current version 2.0.0)
```

Parmetis is trying to rely on the rpath, but the rpath isn't properly being set.  It either needs to set a full path to the parmetis library or correctly reference: `@rpath/libparmetis.dylib` in the library link.

ahmadia:
I'm guessing you're misconfiguring something in CMake from reading this article: http://www.cmake.org/Wiki/CMake_RPATH_handling and inspecting these flags:

```
CMAKE_FLAGS="
  -DCMAKE_VERBOSE_MAKEFILE=1
  -DGKLIB_PATH=${BUILD}/metis/GKlib
  -DCMAKE_INSTALL_PREFIX=${ARTIFACT}
  -DSHARED=1
  -DOPENMP=0
  -DCMAKE_C_COMPILER=$MPICC
  -DCMAKE_CXX_COMPILER=$MPICXX
  -DCMAKE_EXE_LINKER_FLAGS=
  -DCMAKE_INSTALL_RPATH:STRING=${ARTIFACT}/lib
  -DCMAKE_INSTALL_RPATH_USE_LINK_PATH:BOOL=ON"
(
```

ahmadia:
Looks like these flags were put in by @robertmaynard, so he may have some idea of what we're doing wrong here.  I don't see anything in the parmetis build script that seems unreasonable on our part:

```
set -e
export HDIST_IN_BUILD=yes
patch -p1 < _hashdist/no_unused_but_set_variable.patch
export MACOSX_DEPLOYMENT_TARGET=$(sw_vers -productVersion | sed -E "s/([0-9]+\.[0-9]+).*/\1/")
mkdir -p _build_metis
mkdir -p _build_parmetis
CMAKE_FLAGS="
  -DCMAKE_VERBOSE_MAKEFILE=1
  -DGKLIB_PATH=${BUILD}/metis/GKlib
  -DCMAKE_INSTALL_PREFIX=${ARTIFACT}
  -DSHARED=1
  -DOPENMP=0
  -DCMAKE_C_COMPILER=$MPICC
  -DCMAKE_CXX_COMPILER=$MPICXX
  -DCMAKE_EXE_LINKER_FLAGS=
  -DCMAKE_INSTALL_RPATH:STRING=${ARTIFACT}/lib
  -DCMAKE_INSTALL_RPATH_USE_LINK_PATH:BOOL=ON"
(
cd _build_metis && ${CMAKE} ${BUILD}/metis ${CMAKE_FLAGS}
)

(
cd _build_parmetis && ${CMAKE} ${BUILD} ${CMAKE_FLAGS} \
  -DMETIS_PATH=${BUILD}/metis
)
make -C _build_metis -j ${HASHDIST_CPU_COUNT}
make -C _build_parmetis -j ${HASHDIST_CPU_COUNT}
make -C _build_metis install
make -C _build_parmetis install
${MPIEXEC} -np 4 ${ARTIFACT}/bin/ptest ${BUILD}/Graphs/rotor.graph
```

johannesring:
To me it looks like the rpath has been set. When I run `install_name_tool -add_rpath` it tells me that the path already are in `LC_RPATH`:

    [HASHDIST DEBUG] johannr@Johanness-Mac-3:parmetis-evup4xf2ezot-1$ install_name_tool -add_rpath ${ARTIFACT}/lib ${ARTIFACT}/bin/ptest
    error: /Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/install_name_tool: for: /Users/johannr/.hashdist/bld/parmetis/evup4xf2ezot/bin/ptest (for architecture x86_64) option "-add_rpath /Users/johannr/.hashdist/bld/parmetis/evup4xf2ezot/lib" would duplicate path, file already has LC_RPATH for: /Users/johannr/.hashdist/bld/parmetis/evup4xf2ezot/lib

ahmadia:
@johannesring - We'd need to change the name of the installed libraries in each executable from `libparmetis.dylib` to `@rpath/libparmetis.dylib`, the `@rpath/` is implicit in Linux systems but not on OS X's dyld.

ahmadia:
like this:

```
[HASHDIST DEBUG] install_name_tool -change libparmetis.dylib @rpath/libparmetis.dylib ~/.hashdist/bld/parmetis/fqhqgo2cuzxn/bin/ptest
[HASHDIST DEBUG] ~/.hashdist/bld/parmetis/fqhqgo2cuzxn/bin/ptest
RPATH successful expansion of @rpath/libparmetis.dylib to: /Users/aron/.hashdist/bld/parmetis/fqhqgo2cuzxn/lib/libparmetis.dylib
Usage: /Users/aron/.hashdist/bld/parmetis/fqhqgo2cuzxn/bin/ptest <graph-file> [coord-file]
```

ahmadia:
I was just hoping that we could find out what parmetis/CMake is doing wrong here, since it doesn't appear to be our fault.

ahmadia:
What do you want to do?  I think we can just fix the `ptest` binary for now so your tests run on OS X.  Does that seem reasonable?

robertmaynard:
I setup the the CMake flags to enable full rpath so that library is properly loaded independent of the DLYD_LIBRARY_PATH/LD_LIBRARY_PATH setting.

I am curious on what the otool -L output is for libparmetis.dylib and ptest.

If you want to move to using rpath on osx we could change everything over to use MACOSX_RPATH.

johannesring:
Aha, now I understand. Thanks, I'm not too familiar with this on OS X. It would be good to make it work via CMake, but using `install_name_tool` to fix the `ptest` binary is a good alternative. RPath support for OS X was [added](http://www.kitware.com/blog/home/post/510) in CMake 2.8.12 and we are using 2.8.12.2 in HashStack. 

I initially added the test because I had problems getting DOLFIN to build with ParMETIS on OS X.

johannesring:
@robertmaynard - This is the `otool -L` output for `libparmetis.dylib` and `ptest`:

    [HASHDIST DEBUG] johannr@Johanness-Mac-3:parmetis-evup4xf2ezot-1$ otool -L ${ARTIFACT}/lib/libparmetis.dylib
    /Users/johannr/.hashdist/bld/parmetis/evup4xf2ezot/lib/libparmetis.dylib:
    	libparmetis.dylib (compatibility version 0.0.0, current version 0.0.0)
    	/usr/lib/libSystem.B.dylib (compatibility version 1.0.0, current version 1213.0.0)
    	/Users/johannr/.hashdist/bld/mpi/qagb63ozbmab/lib/libmpi.1.dylib (compatibility version 2.0.0, current version 2.8.0)
    [HASHDIST DEBUG] johannr@Johanness-Mac-3:parmetis-evup4xf2ezot-1$ otool -L ${ARTIFACT}/bin/ptest
    /Users/johannr/.hashdist/bld/parmetis/evup4xf2ezot/bin/ptest:
    	libparmetis.dylib (compatibility version 0.0.0, current version 0.0.0)
    	/usr/lib/libSystem.B.dylib (compatibility version 1.0.0, current version 1213.0.0)
    	/Users/johannr/.hashdist/bld/mpi/qagb63ozbmab/lib/libmpi.1.dylib (compatibility version 2.0.0, current version 2.8.0)

johannesring:
Adding `-DCMAKE_MACOSX_RPATH:BOOL=ON` fixed the problem!

ahmadia:
Excellent.

johannesring:
I will update this PR to use `-DCMAKE_MACOSX_RPATH:BOOL=ON`. The `parmetis` package is a special case since it doesn't use the `cmake_package`. Does it matter if we use this flag on all platforms or should we only set it for OS X?

ahmadia:
We should probably only set it on OS X.  It wouldn't be a bad thing to switch it on in `cmake_package.py` as well in the same situation.

ahmadia:
@robertmaynard - do you agree with setting this flag on?  See my earlier comments for a diagnosis of the problem.

robertmaynard:
I don't have lots of experience with CMAKE_MACOSX_RPATH, so I would need to play around with hashstack and verify that the rest of the packages are building with the correct library paths.

ahmadia:
@johannesring - Go ahead and make the change in cmake_package and parmetis and I'll verify against our proteus stack and clawpack.

johannesring:
I have made the change for `parmetis` now and can take a look at the other CMake based packages tomorrow.

BTW, how do you feel about adding tests like this? And what about the `without_check` option to turn it off? Any comments?

ahmadia:
I think you should add these tests everywhere, so long as they are cheap.  It is good to have assurance that things are actually installed correctly.  One problem with this test is that it's more important that the parmetis libraries get installed than the accompanying parmetis binaries.  In general, however, these will help us smoke out problems in our build.

PR #615, opened 2015-02-04
Disable fortran for mpich when the fortran parameter is set to false.

johannesring:


ahmadia:
That looks good, does it work?

johannesring:
Yes, it works. It failed for me first on OS X because I don't have a fortan compiler and the error message told me to add those two flags.

ahmadia:
Thanks @johannesring

ahmadia:
Sounds good.

PR #616, opened 2015-02-04
Update stable to a number of packages

jcftang:
Here's a PR with some updates to the latest stables of various 'helper' packages for what I do.


ahmadia:
Looks good to me, let's wait for build success.

jcftang:
Ok, that sounds good to me.

PR #619, opened 2015-02-06
Add the FORM package

certik:


ahmadia:
This has an unspecified dependency on autotools being installed, but I'm okay with merging it if you need it.

certik:
Let's get it fixed properly. It has:
```
extends: [autotools_package]
```
so it is depending on autotools implicitly. Do you think there is a problem somewhere?

ahmadia:
`autotools_package` means that it uses a `./configure && make && make install` style of build.  Release packages typically provide a pre-built configure script (created by autotools on the packager's machine).  When you're working from a repository, however, you typically need to bootstrap the configure script yourself, using a tool like `autoconf`.  In the build script, you're calling `autoreconf`, which is the full smorgasbord of autotools/automake/libtools builders, so you're implicitly depending on those being available on the hashdist user's system.

ahmadia:
It's not a problem for me, I'm not using this package, but it might be a problem for you :)

jcftang:
You could depend on the autotools that are in hashdist, but it's a lot to be asking for users to install autotools from hashdist to build a package. The autotools packages are a bit fragile at the moment.

certik:
Another package that has the same problem is "mesa".

So what's the solution? Upstream does not provide the configure file, so we need to use our own autotools to rebuild it. @jcftang did you mean 2fd6cfffaeeb3105da7c4c583749880b294b887c is too much to ask, or that without this patch it is too much to ask?

jcftang:
@certik if its just autoconf and or automake I think it would be fine, I think I miss understood that autotools does not include autogen.

certik:
I need to test it on some system without `autoreconf` installed (so far I only tested on my RHEL6 machine that already has `autoreconf` installed), but I think 2fd6cfffaeeb3105da7c4c583749880b294b887c should fix the problem.

certik:
So it turns out I need `autoconf` and `automake`, now things build out of the box. No problem at all.

Except a minor problem on my laptop that autotools itself doesn't build, but I fixed that in #631. The actual local test that I did was that I merged #631 with this PR and only then tested FORM and it works.

Let me know if you have anymore feedback regarding this PR.

ahmadia:
In the future you may also need `libtool`, I believe.

certik:
@ahmadia what do you think might be needed from `libtool`? I uninstalled `libtool` from my Ubuntu and rebuilt FORM --- no problem at all.

Let me know if this PR is ok to merge.

ahmadia:
Yeah, I  just checked this locally.  `libtool` isn't needed by form, it's just something that needed to be explicitly verified in their `configure.ac`.

ahmadia:
This is all good.

certik:
Thanks!

Issue #622, opened 2015-02-11
petsc4py fails to build on OS X - ld: file not found: @rpath/libparmetis.dylib


johannesring:
When using the following profile

    packages:
      launcher:
      python:
        link: shared
      mpi:
        use: mpich
      blas:
        use: host-osx-framework-accelerate
      lapack:
        use: host-osx-framework-accelerate
      petsc:
        build_with: |
          parmetis, scotch, suitesparse, superlu_dist
        coptflags: -O2
        link: shared
        debug: false
      petsc4py:

I get the following error on OS X:

    [petsc4py] /Users/johannr/.hashdist/bld/mpi/gkut75dahgil/bin/mpicc -Wl,-multiply_defined,suppress -Wl,-multiply_defined -Wl,suppress -Wl,-commons,use_dylibs -Wl,-search_paths_first -fPIC -Wall -Wwrite-strings -Wno-strict-aliasing -Wno-unknown-pragmas -O2 -bundle -undefined dynamic_lookup -fno-strict-aliasing -g -O2 -DNDEBUG -g -fwrapv -O3 -Wall -L/Users/johannr/.hashdist/bld/bzip2/gss7yln7fznj/lib -L/Users/johannr/.hashdist/bld/launcher/uroqoaxazfig/lib -L/Users/johannr/.hashdist/bld/ncurses/chfsm2btbjs3/lib -L/Users/johannr/.hashdist/bld/openssl/sfeoc4iaaqjq/lib -L/Users/johannr/.hashdist/bld/readline/bqm6dktmlhfs/lib -L/Users/johannr/.hashdist/bld/sqlite/fofaqjs4h6gx/lib -L/Users/johannr/.hashdist/bld/zlib/6ibn6t2utn27/lib -Wl,-rpath,/Users/johannr/.hashdist/bld/python/cha74esnzltl/lib build/temp.macosx-10.10-x86_64-2.7/src/PETSc.o build/temp.macosx-10.10-x86_64-2.7/src/libpetsc4py.o -L/Users/johannr/.hashdist/bld/petsc/tky7khnxbaml/lib -L/Users/johannr/.hashdist/bld/python/cha74esnzltl/lib -L/Users/johannr/.hashdist/bld/petsc/tky7khnxbaml/lib -lpetsc -o build/lib.macosx-10.10-x86_64-2.7/petsc4py/lib/PETSc.so
[...]

PR #621, opened 2015-02-10
Add superlu_dist

johannesring:

PR #623, opened 2015-02-11
Use PETSC_EXTERNAL_LIB_BASIC when building petsc4py

johannesring:
This will fix issue #622.


ahmadia:
Something isn't right here.  The problem appears to be coming from an unresolved link to a dynamic superlu_dist's library, which you are building as a static library in https://github.com/hashdist/hashstack/pull/621.  I think our build system isn't quite handling this correctly, so I'd prefer to fix this instead of patching petsc4py.  I'm taking a look now.

ahmadia:
Ahh, I see, stupid parmetis.  No this is PETSc's build system's fault.  Let me see if we can massage this a little to do the right thing.

johannesring:
Closed in favor of #626.

PR #626, opened 2015-02-12
Fix install name on parmetis/metis libraries

ahmadia:
This overrides the parmetis/metis library install names to their
absolute paths.  This fixes a number of problems that arise when
downstream packages try to dynamically link them.


ahmadia:
@johannesring - I prefer this to https://github.com/hashdist/hashstack/pull/623.  It should fix https://github.com/hashdist/hashstack/issues/622

johannesring:
This works for me too, so I'm merging it. Thanks.

PR #629, opened 2015-02-12
Bump libadjoint and dolfin-adjoint to 1.5

johannesring:


johannesring:
The build failure is a timeout and not related to this.

PR #631, opened 2015-02-13
Autotools update

certik:
@jcftang you added the initial packages --- any reason why you added such old versions? They don't build anymore on my laptop. This PR fixes it.

The only downside that I can see is that the newest versions are only distributed as `.xz` archives (lzma), which means that `hit` with default Python can't unpack them. I had to install `python-lzma` into my Ubuntu.

I also fully uninstalled any automake/autotools packages, so that I can check that these build (i.e. I discovered the need for 801503f71f41f02e3600b49d043a4276fcfbe20f using this approach).

### DIFF
--- a/pkgs/automake.yaml
+++ b/pkgs/automake.yaml
@@ -1,7 +1,7 @@
 extends: [autotools_package]
 
-#dependencies:
-#  build: [perl]
+dependencies:
+  build: [autoconf]
 
 defaults:
   relocatable: false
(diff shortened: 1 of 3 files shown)


certik:
The automake 1.14 builds on my laptop, but I updated it anyway, while I am at it.

jcftang:
@certik it was a combination of the xz support not quite working for me at the time and also I didn't require the latest versions, just consistently the same versions. Feel free to update!

certik:
Thanks!

PR #638, opened 2015-02-17
Fix the lzma module

certik:
This module is needed for `hit` now, as some packages are only distributed as `.xz` archives.


certik:
I tested this by loading a profile with `lzma` installed and now `hit` works with packages in the `.xz` format.

jcftang:
+1 looks good to me

PR #643, opened 2015-02-17
Append instead of override LDFLAGS and CPPFLAGS

certik:
Now the user can specify LDFLAGS and CPPFLAGS in PROLOGUE and they will not be
overriden later.


ahmadia:
Interesting.  Wouldn't we better off setting "base" LDFLAGS and CPPFLAGS as parameters?  PROLOGUE is a last-ditch workaround when we can't do anything else.

certik:
See here for motivation: https://groups.google.com/d/topic/hashdist/3PvxWc8qa2M/discussion, and then let's see if there is another way to do what was proposed on the list.

certik:
This PR is not needed anymore, as #645 fixes my problem. I don't know if this could be useful for some other uses or not. I am closing it for now, we can reopen later if the need arises.

PR #645, opened 2015-02-18
Make gcc embed rpath automatically

certik:
It embeds RPATH to $ARTIFACT/lib, that way the proper libstdc++ and libgcc will
be used.

Here is "proof" that it works:
```
certik@redhawk:/tmp$ cat a.c
int main()
{
    return 0;
}
certik@redhawk:/tmp$ which g++
/home/certik/repos/hashstack/default/bin/g++
certik@redhawk:/tmp$ g++ a.c
certik@redhawk:/tmp$ ldd a.out 
	linux-vdso.so.1 =>  (0x00007fff12b92000)
	libstdc++.so.6 => /local/certik/bld/gcc/tonmpmygwfax/lib/libstdc++.so.6 (0x00007fee15d67000)
	libm.so.6 => /lib64/libm.so.6 (0x00007fee15ac7000)
	libgcc_s.so.1 => /local/certik/bld/gcc/tonmpmygwfax/lib/libgcc_s.so.1 (0x00007fee158b1000)
	libc.so.6 => /lib64/libc.so.6 (0x00007fee1551d000)
	/lib64/ld-linux-x86-64.so.2 (0x00007fee1607a000)
certik@redhawk:/tmp$ ./a.out 
certik@redhawk:/tmp$ 
```

PR #646, opened 2015-02-19
Make gcc embed rpath automatically

certik:
The patch adds @@ARTIFACT@@ at the proper place in the source code and we then
replace it for ${ARTIFACT} using sed.

### DIFF
--- a/pkgs/gcc/gcc.yaml
+++ b/pkgs/gcc/gcc.yaml
@@ -12,6 +12,19 @@ sources:
   url: http://ftpmirror.gnu.org/gcc/gcc-4.9.2/gcc-4.9.2.tar.bz2
 
 build_stages:
+  - name: rpath_patch
+    before: rpath_sed
+    files: [rpath.patch]
+    handler: bash
+    bash: |
+      patch -up1 < _hashdist/rpath.patch
+
+  - name: rpath_sed
+    before: configure
+    handler: bash
+    bash: |
+      sed -i "s|@@ARTIFACT@@|${ARTIFACT}|g" gcc/config/i386/gnu-user.h gcc/config/i386/gnu-user64.h
+
   - name: link_lib64_to_lib
     after: install
     handler: bash
--- a/pkgs/gcc/gcc.yaml
+++ b/pkgs/gcc/gcc.yaml
@@ -13,16 +13,11 @@ sources:
 
 build_stages:
   - name: rpath_patch
-    before: rpath_sed
+    before: configure
     files: [rpath.patch]
     handler: bash
     bash: |
       patch -up1 < _hashdist/rpath.patch
-
-  - name: rpath_sed
-    before: configure
-    handler: bash
-    bash: |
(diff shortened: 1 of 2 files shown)


pkgs/gcc/gcc.yaml
```diff
@@ -12,6 +12,19 @@ sources:
...
+    files: [rpath.patch]
+    handler: bash
+    bash: |
+      patch -up1 < _hashdist/rpath.patch
+
+  - name: rpath_sed
+    before: configure
+    handler: bash
+    bash: |
+      sed -i "s|@@ARTIFACT@@|${ARTIFACT}|g" gcc/config/i386/gnu-user.h gcc/config/i386/gnu-user64.h
```
certik: I think these two blobs can just be merged into one. I'll do it after my gcc build finishes.

pkgs/gcc/gcc.yaml
```diff
@@ -12,6 +12,14 @@ sources:
   url: http://ftpmirror.gnu.org/gcc/gcc-4.9.2/gcc-4.9.2.tar.bz2
 
 build_stages:
+  - name: rpath_patch
+    before: configure
+    files: [rpath.patch]
+    handler: bash
+    bash: |
+      patch -up1 < _hashdist/rpath.patch
+      sed -i "s|@@ARTIFACT@@|${ARTIFACT}|g" gcc/config/i386/gnu-user.h gcc/config/i386/gnu-user64.h
```
certik: This line assumes that `${ARTIFACT}` does not contain the `|` character. If it does, then the sed command will break.

certik: @ahmadia I've searched the web and there doesn't seem to be any solution really to this, except to manually escape the variable to be substituted. I think it will also fail if the variable contains `[]`. I am inclined to just leave this be (it should work for most paths), unless we can figure out a better solution.

kiwifb: You can use other separator but I would assume that "|" is fairly safe for a path. It is safer than ":" which we could use if there was more than one path.

certik:
I am now running my big stack using gcc built using this PR. So far it works great! I will post once it finishes.

certik:
If all works on my computer, then I would propose to merge this. Somebody with Darwin should then test the patch at https://github.com/hashdist/hashstack/pull/645#issuecomment-74984522 and submit a new PR.

pkgs/gcc/gcc.yaml
certik: Right. The `${ARTIFACT}` path can contain both `:` and `|`, so to be completely safe, one would need to escape it, but it just makes the code really messy.

certik:
Ok, my (huge) stack installed, including the gcc itself (using gcc from this PR), using the gcc from this PR.

This is ready to go in. @ahmadia any objections against merging this?

ahmadia:
Looks good.

certik:
Thanks, merging.

PR #648, opened 2015-02-19
CMake find_path fix for DOLFIN 

johannesring:


johannesring:
Travis error is not related, so merging this old PR.

PR #654, opened 2015-02-20
Add gdbm to default build dependencies for Python

johannesring:
We often require Python to be built with `gdbm`, so it would be nice if this could be the default. What do you think?

### DIFF
--- a/pkgs/python/python.yaml
+++ b/pkgs/python/python.yaml
@@ -20,7 +20,7 @@ when pyver == '3.4':
 
 
 dependencies:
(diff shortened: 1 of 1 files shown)


ahmadia:
No complaints from my side.

certik:
@ahmadia why are we having such downloads problems on Travis? It seems to originate from sourceforge.

certik:
I didn't test this, but I think that's fine.

johannesring:
Thanks. I have restarted the failing tests and will merge this if the they pass.

PR #665, opened 2015-02-22
Remove TAO as additional DOLFIN dependency

cmaurini:
Remove TAO as additional DOLFIN dependency. Since petsc-3.5 TAO is included in PETSC and should be automatically enabled and not considered any more as an additional dependency. The following configuration disable TAO from dolfin by default without any reason.


johannesring:
Thanks.

PR #668, opened 2015-02-22
requirement for  new  jupyter  work

cekees:


certik:
That looks good too.

certik:
Are these needed for the `ipython` git package, or are those needed for some other package, not yet in Hashstack?

cekees:
This one's actually a terminado dependency, so it's actually an optional dependency for ipython.

PR #669, opened 2015-02-23
Add python as mshr dependence

cmaurini:
I think that python should be added as a dependence for `mshr`. Without that, hashdist uses my (old!) system python (2.4 in my case). At least this addition solved my problem. I think that this will solve also this issue: http://fenicsproject.org/pipermail/fenics-support/2015-February/001299.html


certik:
@johannesring, this looks good to me.

johannesring:
Yes, looks good. Thanks.

PR #673, opened 2015-02-23
Fixes gobject-introspection

certik:

PR #671, opened 2015-02-23
Update png, zlib and fix pixman

certik:

PR #674, opened 2015-02-23
[WIP] Pango fixes

certik:


certik:
As of a22f87c29f35e9993facbf72eeaec17e0c304224, it fails with:
```
[pango] checking if g++ supports -c -o file.o... yes
[pango] checking if g++ supports -c -o file.o... (cached) yes
[pango] checking whether the g++ linker (/usr/bin/ld -m elf_x86_64) supports shared libraries... yes
[pango] checking dynamic linker characteristics... (cached) GNU/Linux ld.so
[pango] checking how to hardcode library paths into programs... immediate
[pango] configure: creating ./config.lt
[pango] config.lt: creating libtool
[pango] checking for some Win32 platform... no
[pango] checking for pkg-config... /local/certik/bld/profile/bto56eplrk4v/bin/pkg-config
[pango] checking pkg-config is at least version 0.9.0... 
[pango] yes
[pango] checking for HARFBUZZ... no
[pango] 
[pango] no
[pango] checking for CoreText availability... no
[pango] checking for CAIRO... 
[pango] yes
[pango] checking which cairo font backends could be used... none
[pango] configure: Disabling cairo support
[pango] configure: error: *** Could not enable any backends.
[pango] *** Must have at least one backend to build Pango.
[pango|ERROR] Command '[u'/bin/bash', '_hashdist/build.sh']' returned non-zero exit status 1
[pango|ERROR] command failed (code=1); raising
```
The older version failed with missing `glib.h` library, even though it was specifically included.

jcftang:
I think pango and cairo are circular dependencies.

ahmadia:
Also, please pull the zlib/png updates into separate PRs.  They affect the
entire stack since they're dependencies everywhere, and they are likelier
to be reverted if something goes bad.

certik:
@ahmadia I've already done that in #671, this PR just depends on it. Once #671 is merged, I can rebase this PR.

certik:
@jcftang so we need to bootstrap cairo and pango using the same technique as gettext?

ahmadia:
Great!  Carry on then :)

jcftang:
@certik I think some experimenting might be needed to get the dependencies right, but yes the simple solution would to make it like what gettext does. Do you happen to have cairo already built in the stack that you are testing on right now?

certik:
Yes, I have `cairo` built in the stack, since it fails with `pango`, which requires `cairo`.

@ahmadia, #671 is merged and the #673 contains patches up to 7fcf4f2b85d21dbcfa02b26a0772e03b45657fc2 from this PR (rebased though). After that's in, I'll rebase this PR.

jcftang:
I think if pango and cairo are updated in seperate commits, it is worth merging. It will probably break a few other GTK2 dependencies, but its worth upgrading those as they come along as well.

cekees:
It looks to me like this can go in.

PR #676, opened 2015-02-25
CMB profile

certik:
I took the `cmb-profile` branch from https://github.com/robertmaynard/hashstack/tree/cmb-profile/, rebased it and fixed packages to build.

### DIFF
--- a/examples/cmb.darwin.yaml
+++ b/examples/cmb.darwin.yaml
@@ -1,0 +1,68 @@
+# This profile file controls your <#> (HashDist) build environment.
+
+# In the future, we'll provide better incorporation of
+# automatic environment detection.  For now, have a look
+# at the YAML files in the top-level directory and choose
+# the most *specific* file that matches your environment.
+
+extends:
+ - file: osx.yaml
+
+parameters:
+  HOST_MPICC: /usr/local/bin/mpicc
+  HOST_MPICXX: /usr/local/bin/mpic++
+  HOST_CMAKE: /usr/local/bin/cmake
+  PROLOGUE: |
+    export MACOSX_DEPLOYMENT_TARGET=$(sw_vers -productVersion | sed "s/\(10.[0-9]\).*/\1/"); export CXX=/usr/bin/clang++; export CC=/usr/bin/clang;
+
+# The packages list specifies all the packages that you
+# require installed.  <#> will ensure that all packages
+# and their dependencies are installed when you build this
+# profile.
+
+packages:
+  #blas and lapack are required by numpy
+  blas:
+    use: host-osx-framework-accelerate
+  lapack:
+    use: host-osx-framework-accelerate
+  cmake:
+    use: host-cmake
+  mpi:
+    use: host-mpi
+
+  libxml2:
+  png:
+  szip:
+  zlib:
+  boost:
+    toolset: clang
+    address_model: 64
+
+  #file format packages
+  hdf5:
+  netcdf4:
+  netcdf4cpp:
+
+  #now the rest
+  freetype:
+  gdal:
+  kml:
+  matplotlib:
+  molequeue:
+  numpy:
+  python:
+  qt:
+    toolset: unsupported/macx-clang
+    arch: x86_64
+  remus:
+  shiboken:
+  smtk:
+  vxl:
+  zmq:
+  launcher:
+  paraview:
+  cmb:
+
+  #not worth the time to enable
+  #ffmpeg:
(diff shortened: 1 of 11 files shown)


certik:
@robertmaynard the current status is that all dependencies build on my RHEL6 system (I use gcc 4.9.2 from Hashstack). The CMB package fails with:
```
[cmb] /local/certik/tmp/cmb-psaavgrydzrc/Source/VTKExtensions/General/vtkCmbConeSource.cxx:361:16: error: 'MAXIMUM_NUMBER_OF_PIECES' is not a member of 'vtkStreamingDemandDrivenPipeline'
[cmb]    outInfo->Set(vtkStreamingDemandDrivenPipeline::MAXIMUM_NUMBER_OF_PIECES(),
[cmb]                 ^
```
Which suggests that we have incompatible VTK in Hashstack.

Do you know what exact version of VTK and CMB works together? I am using your original versions, and I can update to the latest versions, unless there needs to be a specific version that works.

robertmaynard:
I believe the ParaView version CMB requires
is 55f3b0ed365b5a14a24ab642b5f5aee0483c1096

certik:
Thanks! I have rebuilt Paraview with 55f3b0ed365b5a14a24ab642b5f5aee0483c1096, but I still get exactly the same error with CMB. I'll investigate what's going on with the `vtkStreamingDemandDrivenPipeline::MAXIMUM_NUMBER_OF_PIECES` symbol in VTK in Paraview.

robertmaynard:
MAXIMUM_NUMBER_OF_PIECES was removed in VTK 6.2, so it looks like you
are linking to a newer version of VTK than the one ParaView was built
against.

certik:
I went into ParaView repository, checked out commit 55f3b0ed365b5a14a24ab642b5f5aee0483c1096, did
```
git submodule init
git submodule update
```
it checks out VTK commit a5daf047a2d480eb465e2dcd50c362674fefe326. `git grep` verifies, that `MAXIMUM_NUMBER_OF_PIECES` is not there. Quick poking around shows that `MAXIMUM_NUMBER_OF_PIECES` was removed in https://github.com/Kitware/VTK/commit/1a0b4e9dde815f84e48ee9912d19be9687c97399.

robertmaynard:
hmmm this is pretty old than. Looks
like e90304c0053a68ab316790748e097d810cb17ed1 would be what you want.

certik:
I used e90304c0053a68ab316790748e097d810cb17ed1 (unfortunately had to put your ParaView patches back in...) and it fixed the problem.

Now I am facing the following problem:
```
[cmb] In file included from /local/certik/tmp/cmb-iumnjqrjr2g5/Source/VTKExtensions/Graphics/vtkCmbArcWidgetRepresentation.cxx:42:0:
[cmb] /local/certik/bld/paraview/kv6q6ywsqz5x/include/paraview-4.1/vtkOpenGL.h:27:43: fatal error: GL/gl.h: No such file or directory
[cmb]  # include <GL/gl.h> // Include OpenGL API.
[cmb]                                            ^
[cmb] compilation terminated.
```
Note that I use mesa, so I use:
```diff
diff --git a/pkgs/cmb.yaml b/pkgs/cmb.yaml
index 97fd653..f3f81ea 100644
--- a/pkgs/cmb.yaml
+++ b/pkgs/cmb.yaml
@@ -3,7 +3,7 @@ dependencies:
   build: [boost, png, qt, zlib, gdal, hdf5,
           mpi, netcdf4, netcdf4cpp, freetype,
           python, numpy, matplotlib, libxml2,
-          paraview, remus, vxl, kml, molequeue, smtk, zmq]
+          paraview, remus, vxl, kml, molequeue, smtk, zmq, mesa]
 
 sources:
 - key: git:b24225a52187a124f5f83bdb7212d363f9254519
```
I need to debug this a bit to see what's going on. ParaView builds just fine.

certik:
As of cdfe64429a806e29668167cc484a728d1b03e5ee, CMB now builds (I used the following profile: https://gist.github.com/certik/570355076297d48438ce).

The problem was fixed by db4a231595dd7cd5580e84ad06f0493a5626908b. @robertmaynard it looks like a bug in CMake or the CMakeLists.txt, see the commit log. I haven't figured out the real cause, just a workaround.

I used the latest CMB commit (b24225a52187a124f5f83bdb7212d363f9254519).

@robertmaynard, how hard would it be to fix CMB, so that it builds against some newer version of ParaView, so that we don't need to ship the temporary patches again? Any other comments to this PR?

certik:
Kitware just opensourced the latest version of CMB, the latest version of dependencies at

https://gitlab.kitware.com/cmb/cmb-superbuild/blob/master/versions.cmake

So now we just need to update this PR.

PR #678, opened 2015-02-26
Allow to set XLIBGL_CFLAGS/LIBS from a profile

### LINKED ISSUE #675: mesa fails to build (configure: error: Package requirements (x11 xext) were not met)
On RHEL6:
```
[mesa] checking for the pthreads library -llthread... no
[mesa] checking whether pthreads work with -pthread... yes
[mesa] checking for joinable pthread attribute... PTHREAD_CREATE_JOINABLE
[mesa] checking if more special flags are required for pthreads... no
[mesa] checking for PTHREAD_PRIO_INHERIT... no
[mesa] configure: Shared GLAPI should not used with Xlib-GLX, disabling
[mesa] checking for LIBDRM... 
[mesa] no
[mesa] checking for LIBUDEV... 
[mesa] no
[mesa] checking for XLIBGL... 
[mesa] configure: error: Package requirements (x11 xext) were not met:
[mesa] 
[mesa] Package x11 was not found in the pkg-config search path.
[mesa] Perhaps you should add the directory containing `x11.pc'
[mesa] to the PKG_CONFIG_PATH environment variable
[mesa] Package 'x11', required by 'world', not found
[mesa] Package 'xext', required by 'world', not found
[mesa] 
[mesa] Consider adjusting the PKG_CONFIG_PATH environment variable if you
[mesa] installed software in a non-standard prefix.
[mesa] 
[mesa] Alternatively, you may set the environment variables XLIBGL_CFLAGS
[mesa] and XLIBGL_LIBS to avoid the need to call pkg-config.
[mesa] See the pkg-config man page for more details.
[mesa] 
[mesa|ERROR] Command '[u'/bin/bash', '_hashdist/build.sh']' returned non-zero exit status 1
[mesa|ERROR] command failed (code=1); raising
```
This patch fixed it for me:
```diff
diff --git a/pkgs/mesa.yaml b/pkgs/mesa.yaml
index 45a159b..0fa7628 100644
--- a/pkgs/mesa.yaml
+++ b/pkgs/mesa.yaml
@@ -1,5 +1,9 @@
 extends: [autotools_package]
 
+dependencies:
+  build: [pkg-config]
+  run: []
+
 sources:
 - key: tar.bz2:a2rdiese5oc4fa7vt5ybmhqg33iqn6bv
   url: ftp://ftp.freedesktop.org/pub/mesa/10.2.4/MesaLib-10.2.4.tar.bz2
@@ -12,6 +16,8 @@ build_stages:
   handler: bash
   bash: |
     autoreconf -fi
+    export XLIBGL_CFLAGS=" "
+    export XLIBGL_LIBS="-lXext -lX11"
 
 - name: configure
   mode: override
```
How should we integrate this to Hashstack? It seems quite platform specific.

certik:
For example on RHEL6, mesa needs the following profile:
```
  mesa:
    xlibgl_cflags: ' '
    xlibgl_libs: '-lXext -lX11'
```
If you build mesa just with the usual profile:
```
  mesa:
```
Then the two environment variables XLIBGL_CFLAGS and XLIBGL_LIBS will be empty,
and they should not affect the build (e.g. if mesa was working before, it
should continue working).

Fixes #675.

### DIFF
--- a/pkgs/mesa.yaml
+++ b/pkgs/mesa.yaml
@@ -1,5 +1,9 @@
 extends: [autotools_package]
 
+dependencies:
+  build: [pkg-config]
+  run: []
+
 sources:
 - key: tar.bz2:a2rdiese5oc4fa7vt5ybmhqg33iqn6bv
   url: ftp://ftp.freedesktop.org/pub/mesa/10.2.4/MesaLib-10.2.4.tar.bz2
@@ -12,6 +16,8 @@ build_stages:
   handler: bash
   bash: |
     autoreconf -fi
+    export XLIBGL_CFLAGS="{{xlibgl_cflags}}"
+    export XLIBGL_LIBS="{{xlibgl_libs}}"
 
 - name: configure
   mode: override


certik:
@ahmadia let me know what you think. This seems to be needed on RHEL6.

ahmadia:
These don't strike me as profile-wide flags, why not just have them set at the package level?

certik:
I thought these are set on the package level. See the commit log (I've added formatting into the issue description).

ahmadia:
Yeah, I see what you mean.  This seems fine.  You could also have libgl export these environment variables when it's a dependency.

certik:
Well `mesa` doesn't have any dependencies, except on the X11 libraries. The `libGl.so` is provided by `mesa`. We could probably introduce `x11` package and just have it taken from the host. I am not an expert though, so I would be happy to have at least this PR merged, so that it works, and we can think if we want to introduce the `x11` package or not.

certik:
@ahmadia can this be merged, or do you have things that have to be done first?

pkgs/mesa.yaml
```diff
@@ -12,6 +16,8 @@ build_stages:
   handler: bash
   bash: |
     autoreconf -fi
+    export XLIBGL_CFLAGS="{{xlibgl_cflags}}"
+    export XLIBGL_LIBS="{{xlibgl_libs}}"
```
ahmadia: In the future this should be an explicit dependency.

ahmadia:
It looks good to merge, feel free to raise an issue to remind yourself to come back and fix the implicit libgl dependency.

certik:
Thanks! Here is the issue: https://github.com/hashdist/hashstack/issues/681

PR #680, opened 2015-03-01
Update Proteus to namespace package

ahmadia:
fix key/url
remove vestigial code


ahmadia:
This allows for private namespace submodules such as proteus-mprans.

pkgs/proteus.yaml
```diff
@@ -0,0 +1,15 @@
...
+
+sources:
+  - key: git:1ee40338ff4a3ac6772c5837228752d6fb62b549
+    url: https://github.com/erdc-cm/proteus
+
+# Proteus is a namespace package, do *not* link the Proteus directory itself in.
+profile_links:
+  - name: python_packages
+    mode: replace
+    link: 'lib/python{{pyver}}/site-packages/proteus/**/*'
```
ahmadia: Folks, this is a useful general trick for installing a namespace package.  It links everything in the proteus directory, but *not* the proteus directory itself.  This allows other packages to link their contents into the proteus directory, which is how several namespace packages install themselves.

certik: Interesting.

PR #683, opened 2015-03-02
fix pelican

ahmadia:
relies on https://github.com/hashdist/hashstack/pull/682

### DIFF
--- a/pkgs/python-markdown.yaml
+++ b/pkgs/python-markdown.yaml
@@ -1,0 +1,9 @@
+extends: [setuptools_package]
+
+dependencies:
+  build: []
+  run: []
+
+sources:
+ - url: https://pypi.python.org/packages/source/M/Markdown/Markdown-2.6.zip
+   key: zip:twdfrwawe3ad52oz37f3kfryfxmqmtrb
--- a/pkgs/typogrify.yaml
+++ b/pkgs/typogrify.yaml
@@ -1,0 +1,9 @@
+extends: [setuptools_package]
+
+dependencies:
+  build: []
+  run: []
+
+sources:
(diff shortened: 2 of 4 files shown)


ahmadia:
Waiting on Travis to double-check that upgrading Jinja doesn't breaking things.

pkgs/python-markdown.yaml
```diff
@@ -0,0 +1,9 @@
+extends: [setuptools_package]
+
+dependencies:
+  build: []
+  run: []
+
+sources:
+ - url: https://pypi.python.org/packages/source/M/Markdown/Markdown-2.6.zip
+   key: zip:twdfrwawe3ad52oz37f3kfryfxmqmtrb
```
certik: flip url and key

pkgs/typogrify.yaml
```diff
@@ -0,0 +1,9 @@
+extends: [setuptools_package]
+
+dependencies:
+  build: []
+  run: []
+
+sources:
+ - url: https://pypi.python.org/packages/source/t/typogrify/typogrify-2.0.7.tar.gz
+   key: tar.gz:rpsgndg2inawhtrctwd4ujz2cgjczmlb
```
certik: flip url and key

certik:
@ahmadia you should just use `hit fetch url` and copy & paste.

ahmadia:
@certik - Those come from `skeleton-pypi`, which has the old order.

ahmadia:
Looks like OpenSSL was down.  My primary concern was Jinja, so I'm merging this.

certik:
@ahmadia then you should flip the url and key, *and* change `skeleton-pypi`. ;)

PR #684, opened 2015-03-02
Update paraview to 4.3.1

certik:
I built it and tested it, it seems to be working.

### DIFF
--- a/pkgs/paraview/paraview.yaml
+++ b/pkgs/paraview/paraview.yaml
@@ -5,7 +5,7 @@ dependencies:
           python, numpy, matplotlib, libxml2, {{build_with}}]
 
 sources:
-- key: git:08cb9fc8d27b24f1d581a98bccd01b8c8b834a19
+- key: git:f7aac037722d9d13b75d9fdb15a1053c0874c6c6
   url: https://github.com/Kitware/ParaView.git
 
 defaults:


certik:
@robertmaynard --- I found a curious thing --- the GMV reader plugin is available in ParaView 4.0.1 in Ubuntu as well as in the version 4.3.1 that this PR introduces. However, for some reason the plugin is not available in ParaView 4.1.0 (currently in Hashstack). Here is the build log of the current Hashstack's paraview on my laptop: https://gist.githubusercontent.com/certik/caed725cc9d2c8e45c69/raw/build.log, it looks like the GMV plugin is being built (`2015/02/27 18:21:44 - INFO: [package:run_job] -- Plugin: GMVReader - Reader for binary or ASCII files stored in General Mesh Viewer file format : Enabled`), but then it is not listed in the list of plugins --- however, ParaView can still load GMV files, except that it can't handle time series (i.e. I have files of the type `data.gmv.00??`).

The Ubuntu version as well as this PR's version seem to work fine for me.

robertmaynard:
ParaView has two GMV Readers, once comes from the VisItBridge and the other from a Plugin. I expect that this is causing the problem with ParaView 4.1.0

certik:
Hm, so on my work machine, both ParaView 4.1.0 and 4.3.1 behave exactly the same: both load GMV as a plugin. On my laptop, I only compiled ParaView 4.1.0 so far, and the plugin is missing there.

certik:
@ahmadia any objections against merging this?

ahmadia:
Go ahead.

certik:
Thanks!

PR #685, opened 2015-03-02
Add Numba

certik:
See commit logs. With this PR, one can execute the script from:

http://matthewrocklin.com/blog/work/2015/02/28/Ising/

I.e.:
```
$ python ising.py 
/home/certik/repos/hashstack/default/lib/python2.7/site-packages/numba/__init__.py:73: UserWarning: llvmlite version format not recognized!
  warnings.warn("llvmlite version format not recognized!")
$
```

### DIFF
--- a/pkgs/enum34.yaml
+++ b/pkgs/enum34.yaml
@@ -1,0 +1,14 @@
+extends: [distutils_package]
+
+dependencies:
+  build: []
+  run: []
+
+sources:
+# The proper sources should be:
+#- key: tar.gz:2paz6jvgundctqmmo5pvtx6f3vmvozgh
+#  url: https://pypi.python.org/packages/source/e/enum34/enum34-1.0.4.tar.gz
+# But they don't work due to https://github.com/hashdist/hashdist/issues/310,
+# so we use this equivalent git repository:
+- key: git:dc35e72a5b4c6da1f0caaab1aaa13553e481e173
+  url: https://github.com/certik/enum34
--- a/pkgs/funcsigs.yaml
+++ b/pkgs/funcsigs.yaml
@@ -1,0 +1,9 @@
+extends: [setuptools_package]
+
+dependencies:
+  build: []
+  run: []
+
+sources:
+- key: tar.gz:3a6onxylb2tgdbya7yo3gu2smoi2ri5n
+  url: https://pypi.python.org/packages/source/f/funcsigs/funcsigs-0.4.tar.gz
--- a/pkgs/llvm.yaml
+++ b/pkgs/llvm.yaml
@@ -6,3 +6,7 @@ sources:
 
 defaults:
   relocatable: false
+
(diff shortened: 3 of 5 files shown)


ahmadia:
Nice, +100 :)

certik:
I also updated Clang and it still builds, so I think this can be merged.

@ahmadia the only possible issue is that I switched to static libraries with LLVM, see the commit log for explanation: 77ca2681da52159b0afa9834d6bf618d76d15a59

ahmadia:
What do the Julia/numba folks need?  I think of them as our two primary LLVM customers :)

certik:
That's right. Numba seems to need static libraries (via `llvmlite`). Julia I thought required dynamic libraries, but given that newer LLVM fails to build and upstream says dynamic are not recommended, we should figure out if Julia can work with static libraries as well.

pitrou:
LLVM breaks its API at each feature release, so you probably want specific versions for each dependent (e.g. llvmlite).

certik:
@pitrou do they break the API when going from 3.5.1 to 3.5.2, or do they only break it when going from 3.5 to 3.6?

@ahmadia, in any case, it looks like we'll have to have llvm3.5, llvm3.4, etc., or maybe even llvm3.5.1, llvm3.5.2 etc. --- essentially we need a particular version for Numba, a particular version for Julia and a particular version for Clang. It should not clash, hopefully.

pitrou:
Le 27/03/2015 16:16, Ondřej Čertík a écrit :
> @pitrou <https://github.com/pitrou> do they break the API when going
> from 3.5.1 to 3.5.2, or do they only break it when going from 3.5 to 3.6?

Only from 3.5 to 3.6.

certik:
@pitrou thanks. Then it's really simple, we just need to have llvm3.4, llvm3.5, etc. The tough part is how to handle Julia, since I think it requires dynamic libraries. Since upstream recommends static, I would make static the default, and for Julia we can create a package llvm3.4-dynamic or something similar.

ahmadia:
@certik - Are we in the situation where one profile needs multiple versions of LLVM simultaneously installed?

certik:
Yes, if you want to try Julia and Numba, from the same profile. But I don't think the profile itself needs to have both LLVM installed, it's just that Julia needs to be built with a different configuration of LLVM than Numba. So our profile variables can't solve the issue, it needs to be a separate package.

ahmadia:
Okay, this needs to be carefully managed, then.  I'd like to avoid
repeating ourselves, where possible, but not sure how to fix this.  @dagss?

dagss:
Me and Mark Florisson discussed this issue in detail, but that's for future features.

So you probably need "base_llvm", then specific "llvm_configured_for_julia" and "llvm_configured_for_numba" packages that extend it to override configuration (hopefully with better names that reflect the configuration difference and not who's using them).

PR #688, opened 2015-03-04
Adding fftw_all.yaml and pyfftw.yaml. 

mikaem:
fftw_all.yaml is a slight modification of original fftw.yaml that builds both single, double and long double versions of the fftw library. All these precision libraries are required by pyfftw, which is a Python wrapper of fftw.

I have only tested this on my own Ubuntu 14.10.

### DIFF
--- a/pkgs/fftw_all.yaml
+++ b/pkgs/fftw_all.yaml
@@ -1,0 +1,36 @@
+extends: [autotools_package]
+dependencies:
+  build: [mpi]
+
+sources:
+- url: http://www.fftw.org/fftw-3.3.3.tar.gz
+  key: tar.gz:qxg7ycqluegy7jhq7dtthkwbuwjwzbmy
+
+defaults:
+  # lib/libfftw3.la contains hard-coded path
+  relocatable: false
+
+build_stages:
+- name: configure
+  extra: ['--disable-debug', '--enable-shared', '--enable-mpi',
+          '--enable-sse2', '--enable-avx', '--enable-threads']
+
+- name: install
+  after: make
+  handler: bash
+  mode: override
+  bash: |
+    make install
+    make clean
+    # Build single precision library as well:
+    ./configure --prefix=$ARTIFACT --enable-single --disable-debug \
+            --enable-shared --enable-mpi --enable-sse2 --enable-avx \
+            --enable-threads
+    make -j ${HASHDIST_CPU_COUNT}
+    make install
+    make clean
+    # Build long double precision library as well:
+    ./configure --prefix=$ARTIFACT --enable-long-double --disable-debug \
+            --enable-shared --enable-mpi --enable-threads
+    make -j ${HASHDIST_CPU_COUNT}
+    make install
--- a/pkgs/pyfftw.yaml
+++ b/pkgs/pyfftw.yaml
@@ -1,0 +1,17 @@
+extends: [distutils_package]
+
+dependencies:
+  build: [fftw, mpi4py, mpi, numpy, cython]
+  run: [numpy]
+
+sources:
+- key: tar.gz:6253nl5jgccubgvsjcc2di6nxcij6ck2
+  url: https://pypi.python.org/packages/source/p/pyFFTW/pyFFTW-0.9.2.tar.gz
+
+build_stages:
+- name: install
+  after: setup_dirs
+  handler: bash
+  bash: |
+    ${PYTHON} setup.py build_ext --inplace --library-dirs=$ARTIFACT/lib --include-dirs=$ARTIFACT/include
+    ${PYTHON} setup.py install --prefix=$ARTIFACT


certik:
Why not to just add your modifications to the `fftw.yaml`?

certik:
Otherwise it looks good, thanks for the PR!

ahmadia:
Thanks for the PR @mikaem.  I think we may be able to fold this variant into fftw.yaml.  Would that be acceptable to you?

mikaem:
For me it makes no difference, just didn't want to change what was put there originally and long double precision is normally not needed. Is there a clever way of modifying fftw.yaml such that one may choose precision?

ahmadia:
>  Is there a clever way of modifying fftw.yaml such that one may choose precision?

I'm glad you asked :)  This is a great place to use a package parameter and `when` clauses.  There are some good examples in https://github.com/hashdist/hashstack/blob/master/pkgs/python/python.yaml#L3.  A corresponding profile example is here: https://github.com/hashdist/hashstack/blob/master/tests/test.scipy-stack.yaml#L8

I can take a look at this later if it looks like you're stuck.

certik:
I agree with @ahmadia. Plus, if long double is needed for pyfftw, then it is important to have it.

mikaem:
Thanks for the feedback. I'll dig a little and see if I can figure out how to set this with options in profiles.

mikaem:
Ok, this is my first effort. If I put this parameter in my profile yaml

parameters:
  fftw_precision: 'sd'   # options 'sd' or 'sdl'

then I can write in fftw.yaml:

build_stages:
  - name: configure
    extra: ['--disable-debug', '--enable-shared', '--enable-mpi',
            '--enable-sse2', '--enable-avx', '--enable-threads']

  - when: fftw_precision == 'sd'
    name: install
    after: make
    handler: bash
    mode: override
    bash: |
      make install
      make clean
      # Build single precision library as well:
      ./configure --prefix=$ARTIFACT --enable-single --disable-debug \
            --enable-shared --enable-mpi --enable-sse2 --enable-avx \
            --enable-threads
      make -j ${HASHDIST_CPU_COUNT}
      make install
    
  - when: fftw_precision == 'sdl'    
    name: install
    after: make
    handler: bash
    mode: override
    bash: |
      make install
      make clean
      # Build single precision library as well:
      ./configure --prefix=$ARTIFACT --enable-single --disable-debug \
            --enable-shared --enable-mpi --enable-sse2 --enable-avx \
            --enable-threads
      make -j ${HASHDIST_CPU_COUNT}
      make install
      make clean
      # Build long double precision library as well:
      ./configure --prefix=$ARTIFACT --enable-long-double --disable-debug \
            --enable-shared --enable-mpi --enable-threads
      make -j ${HASHDIST_CPU_COUNT}
      make install

Not very pretty, but it seems to work. Any better ideas? I don't really know any yaml at all, and my next idea would probably be to generate the file using a small python script.

ahmadia:
Yes, that's exactly what I had in mind!

As a further refinement, you could have `enable_single` and `enable_long_double` as top-level parameters in the profile file.  These could be general triggers (not just for FFTW) that the profile should contain single-precision or double-precision variants.

Also, I didn't point this out with a link, but the [HashDist YAML Domain Specific Language](http://hashdist.readthedocs.org/en/latest/specs.html#stage-system) can simplify your logic using the `update` flag to the configure stage.  This is functionality we only added a couple months ago, so it's fresh!

```yaml

defaults:
   enable_single: false
   enable_long_double: false

build_stages:
  - name: configure
    extra: ['--disable-debug', '--enable-shared', '--enable-mpi',
            '--enable-sse2', '--enable-avx', '--enable-threads']

  - when: enable_single
    name: configure
    mode: update
    extra: ['--enable-single']

  - when: enable_long_double
    name: configure
    mode: update
    extra: ['--enable-long-double']
```

Due to the we make use of CSS-like "inheritance" in HashDist, this should do what you want, and hopefully be easier for us to maintain in the long run.  I think the use of `enable_single` and `enable_long_double` is something we can endeavor to support in a uniform manner across HashStack.

mikaem:
So, if I enable both single and long double, then you are saying that all 3 precisions of the libraries will be built? Btw, long double cannot have --enable-sse2 or --enable-avx, so these must be removed for the long double configure. What's the opposite of extra?

ahmadia:
> So, if I enable both single and long double, then you are saying that all 3 precisions of the libraries will be built?

That's my understanding of how FFTW's build system works, but it's been awhile since I looked at it.

> Btw, long double cannot have --enable-sse2 or --enable-avx, so these must be removed for the long double configure. What's the opposite of extra?

You would have to remove the update, we don't currently have a way to smartly prune from `extra`.  Are you sure `--enable-sse2` and `--enable-avx` are incompatible?  I just assume they wouldn't be used for the long-double versions of the library.

mikaem:
| That's my understanding of how FFTW's build system works, but it's been awhile since I looked at it.

You have to configure and make install the double precision, then make clean; make install with enable-single and then make clean; make install with enable-long.

I got error messages for the long double on my machine (ubuntu 14.10) and had to remove enable-sse2 and enable-avx. Then it worked.

ahmadia:
Got it.  Let me take a look at your proposal again, I think you had the right way of it.

ahmadia:
Okay, what I don't like your about your approach is that we lose the global configure options when we build the single and long-double precision.  If we have to "reconfigure" though, there aren't going to be any great options.  Some ideas that might work:

* break things up into fftw_single, fftw (standard double-precision), fftw_long_double, and just build that variant for each

* use the fftw for regular double-precision, fftw_all to build all variants

* enable our fftw.yaml file to build the other variants.  This is messy, though, because there is no good way to re-run the configure stage in HashDist.

I'm fine with your original proposal (`fftw_all`), or breaking things up into multiple files.  I didn't realize the FFTW build system was so futzy or I would have green-lit your approach from the start.  Sorry!

@certik - what do you think?

jcftang:
So are we sticking with the current/last proposal or going back to the initial one for this PR?

cekees:
I'm +1 on going back to the original approach or just putting these in fftw.

ahmadia:
I'm fine with merging these in.  @certik?

certik:
This PR doesn't break anything, so we can merge it later if we figure out a nice way.

PR #692, opened 2015-03-06
libvterm: Initial add

jcftang:


pkgs/libvterm.yaml
```diff
@@ -0,0 +1,16 @@
+extends: [base_package]
+
+sources:
+- key: git:20ad1396c178c72873aeeb2870bd726f847acb70
+  url: https://github.com/neovim/libvterm.git
+
+build_stages:
+- name: make_and_install
+  handler: bash
```
certik: You should add `after: prologue` to ensure that user defined PROLOGUE is executed first.

certik:
Otherwise +1

jcftang:
Fixed.

PR #697, opened 2015-03-13
Update cbcblock.yaml

MiroK:
The previous version was using an old PETSc interface and this resulted in error `AttributeError: type object 'MatStructure' has no attribute 'SAME_PRECONDITIONER’` whenever ML preconditioner was constructed. Yaml now points to the lastest commit in cbc.block where the
error is fixed


johannesring:
Thanks!

certik:
Thanks @MiroK for the PR.

ahmadia:
Thanks @MiroK.

PR #702, opened 2015-03-25
Add libxml2 and pkg-config to build deps for mpich

johannesring:


certik:
That looks good. I didn't know those were dependencies.

jcftang:
It's probably needed for the trace file generation, mpelog if i remember right

jcftang:
lets merge this, I think it's reasonable to add this dependency

PR #703, opened 2015-03-25
Add Libint

certik:


pkgs/libint.yaml
```diff
@@ -0,0 +1,25 @@
...
+  build: [autoconf, automake, gmp, boost]
+
+sources:
+# When https://github.com/evaleev/libint/pull/30 is merged, we can switch to
+# use the original repository, until then we use a clone with the necessary
+# patches:
+#- key: git:c1be4f680ada1797e71741fa882b68e926fc05a9
+#  url: https://github.com/evaleev/libint
+- key: git:c0a83fa7e2554ea71a14dbae47bce0a3559e1a54
+  url: https://github.com/certik/libint-1
```
certik: @ahmadia any comments on this? Is it ok to use my own github repo, or should we rather use the official repo and add patches to Hashstack?

jcftang:
I think using the upstream repo and adding patches is better that pointing to personal repos. Otherwise +1.

certik:
@jcftang ok. Upstream merged my patch, so I am just using upstream now.

jcftang:
+1

PR #720, opened 2015-04-09
Trilinos update

certik:
See the commit logs. Now one can switch to git version of Trilinos just by changing the sources, and e.g. the latest master builds without changing anything else.


johannesring:
+1, looks good.

PR #722, opened 2015-04-11
Update Python

vbraun:
Just a minor point release...

### DIFF
--- a/pkgs/python/python.yaml
+++ b/pkgs/python/python.yaml
@@ -2,8 +2,8 @@ extends: [autotools_package]
 
 when pyver == '2.7':
     sources:
-    - key: tar.gz:otlqxeknurehviozoirlfhuvktief6bf
-      url: http://python.org/ftp/python/2.7.8/Python-2.7.8.tgz
+    - key: tar.gz:zc52gptgvqzadwv5yvlpb2t47zvmcgkg
+      url: http://python.org/ftp/python/2.7.9/Python-2.7.9.tgz
 when pyver == '3.2':
     sources:
     - key: tar.gz:7qpecklofhkhn5uwga5mvyutvz5cgehq
@@ -14,8 +14,8 @@ when pyver == '3.3':
       url: https://www.python.org/ftp/python/3.3.6/Python-3.3.6.tgz
 when pyver == '3.4':
     sources:
-    - key: tar.gz:isr4d3y4psr6j7jfeqvpqdwxfwuucib4
-      url: https://www.python.org/ftp/python/3.4.2/Python-3.4.2.tgz
+    - key: tar.gz:rn2d6vxj4uf7ber3t2oelxmspqdr26vf
+      url: https://www.python.org/ftp/python/3.4.3/Python-3.4.3.tgz
 
 
 


vbraun:
Mhh the certificate of sqlite.org expired, so some tests fail.

certik:
```
Downloading https://sqlite.org/2013/sqlite-autoconf-3071700.tar.gz...
[ERROR] urllib failed to download (reason: [SSL: CERTIFICATE_VERIFY_FAILED] certificate verify failed (_ssl.c:581)): https://sqlite.org/2013/sqlite-autoconf-3071700.tar.gz
```
What should we do about this? Should we turn off certificate checking in hit? We still check the hash of the downloaded tarball, so this should be safe.

Ondrej

vbraun:
But `hit fetch` doesn't check the certificate on my machine. Or, at the very least, uses a different set of certificates to verify. If the travis error is from hit then it does check the cert there, so it might all boil down to different certificates.

Also, today I can visit https://sqlite.org in chrome and it works (didn't work a 2 days ago), so maybe the certificate issue is being sorted out right now.

vbraun:
I investigated a bit and
* sqlite.org just replaced their expired certificate (good)
* sqlite.org removed the intermediate certificates when updating (bad)

Hence TLS still fails depending on whether your client includes intermediate (non-root) certificates or not.

vbraun:
upstream mailinglist thread: http://thread.gmane.org/gmane.comp.version-control.fossil-scm.user/20821

ahmadia:
@vbraun - Thanks for following up on this.  I'd like to, where possible, isolate ourselves from upstream errors.  One option is to provide an offsite source cache for tarballs that we know have temporary download problems.  I can stand something up relatively quickly.  Does that work for everybody?

vbraun:
The sqlite TLS issue seems to be fixed  now!

A good system to manage tarballs-by-checksum would be much appreciated. I'm sick and tired of copying tarballs by hand from google drive shares to the sagemath mirrors ;-)  The main design questions are IMHO:   
* Authentication should be minimal, so that as many different people as possible can upload tarballs (minimize administration).
* Anonymous file uploads shouldn't be downloadable by default, or it will be abused to host priacy/malware. Unless its otherwise too constrained to be useful (bandwidth/size quota)

certik:
How about checking the certificate for `hit fetch`, but not for `hit build` (because for `hit build` we already have a hash, that should be cryptographically secure).

ahmadia:
> A good system to manage tarballs-by-checksum would be much appreciated. I'm sick and tired of copying tarballs by hand from google drive shares to the sagemath mirrors ;-) 

@vbraun - Do you know how to point to remote source caches in a HashDist install?

vbraun:
How is that going to help, people that don't have a webserver to host files aren't going to host their own source cache either. Also, if every individual developer hosts their own files then you are back to relying on the weakest link in the chain.

ahmadia:
I think I misunderstood your request.  My idea was to maintain a web store with a list of source tarballs needed by Travis to run its tests.  We could give Travis permission to upload tarballs that its used back to this server.  I don't think this would be hard to install and maintain.

ahmadia:
This looks good, by the way.

PR #725, opened 2015-04-12
Add coffeescript

vbraun:
This is a bit of an experiment in packaging node.js modules. Since there is basically no stable node.js module, the node package manager makes private installs of all dependencies for each package (recursively).

However, the coffeescript transpiler has no dependencies so its actually easy ;-)

### DIFF
--- a/pkgs/coffeescript.yaml
+++ b/pkgs/coffeescript.yaml
@@ -1,0 +1,20 @@
+extends: [base_package]
+dependencies:
+  build: [node]
+
+defaults:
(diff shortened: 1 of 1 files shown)


ahmadia:
That seems reasonable.  I guess we will eventually want a node package installer.  @jcftang - any comments?

jcftang:
Looks good to me. +1

ahmadia:
Thanks.

certik:
Looks reasonable, thanks!

PR #728, opened 2015-04-15
netcdf: Rebase the patch on top of the latest code

certik:
Previously it failed to apply (due to white space), now it works again.


certik:
This works for me, so I am merging this.

PR #732, opened 2015-04-29
hypre rpath-fix

JaroslavHron:
The rpath-fix should be applied to the real library not to the symlink.
If applied to the symlink (i.e. libHYPRE.so), the symlink is changed to a copy of the file and the rpath of the original library (i.e. libHYPRE-x.x.x.so)  file is not fixed.


certik:
I think that's fine, if it fixes it.

johannesring:
Thanks!

Issue #696, opened 2015-03-12
numpy isn't linked to any libraries as default. 


terjekv:
After installing numpy via hashstack, the end result is somewhat disappointing:

$ module load FEniCS/1.5.0_2015.02.20_dev
$ python
imPython 2.7.8 (default, Feb 20 2015, 22:02:57) 
[GCC 4.8.2 20140120 (Red Hat 4.8.2-16)] on linux2
Type "help", "copyright", "credits" or "license" for more information.
po>>> import numpy
>>> numpy.__config__.show()
lapack_info:
  NOT AVAILABLE
lapack_opt_info:
  NOT AVAILABLE
blas_info:
  NOT AVAILABLE
atlas_threads_info:
  NOT AVAILABLE
blas_src_info:
  NOT AVAILABLE
atlas_blas_info:
  NOT AVAILABLE
lapack_src_info:
  NOT AVAILABLE
openblas_info:
  NOT AVAILABLE
atlas_blas_threads_info:
  NOT AVAILABLE
blas_mkl_info:
  NOT AVAILABLE
blas_opt_info:
  NOT AVAILABLE
atlas_info:
  NOT AVAILABLE
lapack_mkl_info:
  NOT AVAILABLE
mkl_info:
  NOT AVAILABLE

This is solvable by editing the yaml file to include the following:

build_stages:

- when: platform == 'linux'
  name: create-site.cfg
  before: install
  handler: bash
  bash: |
    cat > site.cfg << EOF
    [blas]
    blas_libs = blas
    library_dirs = ${ARTIFACT}/lib
    [lapack]
    lapack_libs = lapack
    library_dirs = ${ARTIFACT}/lib
    EOF

But I'm not sure if that's a pretty solution.

ahmadia:
Hmnn.  So you'd like NumPy to use the LAPACK libraries in the same way that SciPy does on Linux?  This seems reasonable to me.  @certik?

certik:
I think something along these lines. The current numpy yaml is clearly broken:
```
extends: [distutils_package]
dependencies:
  build: [blas]
  run: [blas]

sources:
  - key: tar.gz:zzluhi5cjpyt4a3t7lvsicyi4sgrhhtd
    url: http://downloads.sourceforge.net/numpy/numpy-1.8.2.tar.gz

build_stages:

  - when: platform == 'linux'
    name: create-site.cfg
    before: install
    handler: bash
    bash: |
      cat > site.cfg << EOF
      [atlas]
[...]

PR #733, opened 2015-04-29
scipy - add better openblas configuration and umfpack/suitesparse dependency.

JaroslavHron:
This sets the openblas as optimized blas and lapack for scipy and adds depndency on suitesparse (amd/umfpack)

### DIFF
--- a/pkgs/scipy/scipy.yaml
+++ b/pkgs/scipy/scipy.yaml
@@ -3,7 +3,7 @@ when platform != 'linux':
 when platform == 'linux':
   extends: [distutils_package, libflags]
 dependencies:
-  build: [lapack, numpy]
+  build: [lapack, numpy, suitesparse]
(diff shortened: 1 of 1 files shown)


certik:
I think this will fix it to work with openblas, but I don't think this will work if I want to use a different lapack. So this is not the way.

JaroslavHron:
I see - after reading the discussions https://github.com/hashdist/hashstack/issues/696 and https://github.com/hashdist/hashstack/pull/709 its clear some more thought is needed on blas/lapack...
This is at least for me working solution with openblas inside hashdist.

certik:
Indeed, I was just looking where I posted my comment about this: https://github.com/hashdist/hashstack/pull/709#issuecomment-88595480. Let's keep this PR open, to remind us to fix this once and for good, and also because this is a quick solution if you want to use openblas.

PR #734, opened 2015-04-29
numpy - add better openblas configuration.

JaroslavHron:
Sets openblas as optimized blas/lapack for numpy.


JaroslavHron:
Closed for the same reason as https://github.com/hashdist/hashstack/pull/733

Issue #727, opened 2015-04-13
gcc fails to build on OS X 10.10 (Yosemite)


certik:
```
[gcc] 		-DLIBGCC2_UNITS_PER_WORD=4
[gcc] /Users/certik/.hashdist/tmp/gcc-l7u34nnvgab4/host-x86_64-apple-darwin14.1.0/gcc/xgcc -B/Users/certik/.hashdist/tmp/gcc-l7u34nnvgab4/host-x86_64-apple-darwin14.1.0/gcc/ -B/Users/certik/.hashdist/bld/gcc/l7u34nnvgab4/x86_64-apple-darwin14.1.0/bin/ -B/Users/certik/.hashdist/bld/gcc/l7u34nnvgab4/x86_64-apple-darwin14.1.0/lib/ -isystem /Users/certik/.hashdist/bld/gcc/l7u34nnvgab4/x86_64-apple-darwin14.1.0/include -isystem /Users/certik/.hashdist/bld/gcc/l7u34nnvgab4/x86_64-apple-darwin14.1.0/sys-include    -g -O2 -m32 -O2  -g -O2 -DIN_GCC    -W -Wall -Wno-narrowing -Wwrite-strings -Wcast-qual -Wno-format -Wstrict-prototypes -Wmissing-prototypes -Wold-style-definition  -isystem ./include   -pipe -fno-common -g -DIN_LIBGCC2 -fbuilding-libgcc -fno-stack-protector   -pipe -fno-common -I. -I. -I../../../host-x86_64-apple-darwin14.1.0/gcc -I../../.././libgcc -I../../.././libgcc/. -I../../.././libgcc/../gcc -I../../.././libgcc/../include  -DHAVE_CC_TLS -DUSE_EMUTLS -o _fixunsdfsi.o -MT _fixunsdfsi.o -MD -MP -MF _fixunsdfsi.dep -DL_fixunsdfsi -c ../../.././libgcc/libgcc2.c -fvisibility=hidden -DHIDE_EXPORTS \
[gcc] 		-DLIBGCC2_UNITS_PER_WORD=4
[gcc] ../../.././libgcc/libgcc2.c: In function '__gcc_bcmp':
[gcc] ../../.././libgcc/libgcc2.c:2105:1: internal compiler error: Segmentation fault: 11
[gcc]  }
[gcc]  ^
[gcc] libbacktrace could not find executable to open
[gcc] Please submit a full bug report,
[gcc] with preprocessed source if appropriate.
[...]

PR #735, opened 2015-04-30
Make GCC build on OS X 10.10

certik:


certik:
As of 11b1283e1fc2260ac9aee1ea49b1d7188492e837, this builds gcc.

certik:
Also 7c963e8b2d2511f67fe83f874f1c41daaf498678 builds.

certik:
It looks like it is not about these dependencies --- it can pick them up from Hashdist just fine (it can probably pick gmp as well). It looks like it is about setting the path to that profile during gcc build --- either the `-I` or `-L` flags. This would suggest it requires some dependency from my `default` profile, and if it doesn't have it, one gets internal compiler error. Here is my default profile that I've been using with this PR:
```
extends:
- file: osx_clt.yaml

packages:
  launcher:
  cmake:
  python:
    host: true
  blas:
    use: host-osx-framework-accelerate
  lapack:
    use: host-osx-framework-accelerate
  mpi:
    use: openmpi
  boost:
    toolset: clang
    address_model: 64
  ipython:
  # gcc deps:
  zlib:
  mpfr:
  mpc:
  gmp:
  isl:
  cloog:
  gnu-sed:
```

certik:
Interesting is that 0b89fb8a2e899812639037a73c21ce532a517f19 fails (as before), so perhaps it has something to do with gmp after all.

certik:
Also builds with 95333e89aa8bb288bf0a9e71b679da5f94756dd4.

certik:
This PR helped bisecting it. I then reduced the profile that I was using for gmp, until only gmp was in it. I then compared all possible differences, and I only spotted two, the `MAC_DEPLOYMENT_TARGET` and that `PATH` had `/usr/local/bin` in it (that directory does not exist on my machine). I then tried unsetting the `MAC_DEPLOYMENT_TARGET` for gmp and that did it.

See https://github.com/hashdist/hashstack/issues/727#issuecomment-97922225 for more details.

PR #737, opened 2015-04-30
httpie: initial add of package, this is a curl alternative

jcftang:
This is a curl alternative and is very useful for testing against web based api's


certik:
+1

jcftang:
@certik thanks!

Issue #741, opened 2015-05-01
Working GCC profile on OS X 10.10


certik:
Using the latest master (0e6de248224a2e340ec6f8ba9f10ef90d7a87788), and the default Xcode (installed automatically by OS X if you first type `gcc` into a terminal), this profile builds (among other things, gcc):
```
$ cat hd_base.yaml 
extends:
 - file: osx.yaml

packages:

  launcher:
  blas:
    use: host-osx-framework-accelerate
  lapack:
    use: host-osx-framework-accelerate
  mpi:
    use: mpich
  python:
    link: shared
  swig:
    build_with: |
      perl
  ipython:
  nose:
  numpy:
  sphinx:
  sympy:
  gmp:
  matplotlib:
  doxygen:
  breathe:
  gcc:
  git:
  perl:
  pyliblzma:
```
Once the `hd_base` profile is built, obtain it's path by:
```
$ ll hd_base
lrwxr-xr-x  1 certik  staff  48 May  1 09:46 hd_base@ -> /Users/certik/.hashdist/bld/profile/px524wdzacsz
```
and then use the following profile to build Hashstack (use the path from above in `PROLOGUE`):
```
extends:
 - file: osx.yaml

parameters:
  # First line is copied from osx.yaml (we cannot append to PROLOGUE yet)
  PROLOGUE: |
    export MACOSX_DEPLOYMENT_TARGET=$(sw_vers -productVersion | sed -E "s/([0-9]+\.[0-9]+).*/\1/"); export PATH="/Users/certik/.hashdist/bld/profile/px524wdzacsz/bin:$PATH"; export CC=gcc CXX=g++ FC=gfortran

packages:

  launcher:
  blas:
    use: host-osx-framework-accelerate
  lapack:
    use: host-osx-framework-accelerate
  mpi:
    use: mpich
  python:
    link: shared
  swig:
    build_with: |
      perl
  ipython:
  nose:
  numpy:
  sphinx:
  sympy:
  gmp:
  doxygen:
  breathe:
  perl:
  pyliblzma:
  git:
    disable_crypto: true
  csympy:
```
You need #743 in order for this to build. Then it builds nicely.

I will keep updating this issue with the latest progress. Eventually we'll make it an example in our examples directory and close this.

certik:
Also, one really needs to also enable `fortran: true` in the second profile, otherwise `mpif90` isn't built and other packages like `scalapack` fail.

[...]

PR #743, opened 2015-05-01
[WIP] Fixes for OS X

certik:
I should probably split this into several PRs, but I am posting it here so that others can use it already.

It also needs to be tested on linux.


pkgs/gcc/gcc.yaml
```diff
@@ -22,6 +22,14 @@ build_stages:
       patch -up1 < _hashdist/rpath.patch
       sed -i "s|@@ARTIFACT@@|${ARTIFACT}|g" gcc/config/i386/gnu-user.h gcc/config/i386/gnu-user64.h
 
+  - when: platform == 'Darwin'
+    name: fix_object_h
+    before: configure
+    handler: bash
+    bash: |
+      mkdir -p "$ARTIFACT/include/dispatch"
+      sed 's+typedef void (\^dispatch_block_t)(void)+typedef void* dispatch_block_t+' /usr/include/dispatch/object.h > "$ARTIFACT/include/dispatch/object.h"
```
certik: I think I should add a check that /usr/include/dispatch/object.h exists in the first place, and then also do this fix for OS X 10.10 only.

pkgs/git.yaml
```diff
@@ -19,6 +20,13 @@ build_stages:
   bash: |
     export PERL_PATH="/usr/bin/env perl"
 
+- when: platform == 'Darwin' and disable_crypto
+  name: fix_crypto_compile_failure
+  before: configure
+  handler: bash
+  bash: |
+    export NO_APPLE_COMMON_CRYPTO=1
```
certik: This disables https support, so it is not a solution. It makes `git` build with our gcc at least, so that's why I did it, but we need to find a real solution.

certik:
@ahmadia, @cekees these are changes that I needed to make to get at least some initial profile working with Hashdist's gcc (see #741).

Many packages still fail to build. The above must work, so we just need to figure out the proper fixes to some of the hacks that I had to do in this PR. I would appreciate any feedback or help with this.

I am new to OS X. How do you use Hashdist on OS X? Do you just use the native Apple clang? If so, how do you compile Fortran packages? Any alternative way to use Hashdist must also work.

jcftang:
@certik I think if the updates are logical commits its fine for it to be in one merge

certik:
I have split this PR into #778, #774, #773, #772, #771 and #769. That way we can work on improving some of the fixes, which are not ideal, and merge those that are ok. Closing.

PR #751, opened 2015-05-02
update IPython to 3.1

minrk:
- add missing dependency on markupsafe to jinja2
- add missing dependency on jsonschema to ipython
- update tornado to 4.1 (required for IPython 3)
- add backports.ssl_match_hostname (required for tornado 4)

### DIFF
--- a/pkgs/backports-ssl_match_hostname.yaml
+++ b/pkgs/backports-ssl_match_hostname.yaml
@@ -1,9 +1,5 @@
 extends: [setuptools_package]
 
-dependencies:
-  build: []
-  run: []
-
 sources:
 - key: tar.gz:a5aq475qtkvxxwxv4ymn4zwd3lee4lr5
   url: https://pypi.python.org/packages/source/b/backports.ssl_match_hostname/backports.ssl_match_hostname-3.4.0.2.tar.gz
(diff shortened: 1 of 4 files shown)


certik:
Tests failed for jinja2. It needs to be investigated.

minrk:
jinja2 yaml seems to have needed a newline that other yaml files don't have. Weird.

ahmadia:
Hi, a bit late to the party, but a bunch of this work was already done in the private ROAMS stack...

certik:
I don't know what ROAMS is, but you should opensource more of your private work. :)

cekees:
We're trying... The overlap here  is not quite duplication. What minrk is
pushing is for the actual IPython 3.1.0 release whereas the private
 repository was based on dev branches and forks of the official work that
wasn't ready for the public hashstack.  I'm working to minimize the number
of packages that aren't coming directly from hashstack master or at  least
a branch of hashstack.

cekees:
I've tested this on ubuntu on several HPC machines. I'm fine with it going into master.

pkgs/backports-ssl_match_hostname.yaml
```diff
@@ -0,0 +1,9 @@
+extends: [setuptools_package]
+
+dependencies:
+  build: []
+  run: []
```
certik: This `dependencies` section can be removed.

certik:
Only a minor comment above. +1 to merge.

pkgs/backports-ssl_match_hostname.yaml
minrk: removed

certik:
Thanks!

PR #754, opened 2015-05-06
Add Ninja

certik:
This can be used with `cmake`, just pass `-G Ninja` to it. Works great.

### DIFF
--- a/pkgs/ninja.yaml
+++ b/pkgs/ninja.yaml
@@ -4,8 +4,8 @@ dependencies:
   build: [python]
 
 sources:
-- key: git:3309498174411e02e7680ea8b470bb7d1d70bdb8
-  url: https://github.com/martine/ninja
+- key: tar.gz:psktwwt4e3h42cbiqlr7hywnbd7orbek
+  url: https://github.com/martine/ninja/archive/v1.5.3.tar.gz
 
 build_stages:
 - name: build
--- a/pkgs/ninja.yaml
+++ b/pkgs/ninja.yaml
@@ -1,0 +1,17 @@
+extends: [base_package]
+
+dependencies:
+  build: [python]
+
+sources:
+- key: git:3309498174411e02e7680ea8b470bb7d1d70bdb8
+  url: https://github.com/martine/ninja
+
+build_stages:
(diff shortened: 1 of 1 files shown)


ahmadia:
Any particular reason you didn't point to a release?

certik:
I did.

ahmadia:
Ahh, the Git repository is the slowest way to get sources, since the entire Git repository is cloned.  You're better off pointing to a tarball like this one: https://github.com/martine/ninja/archive/v1.5.3.tar.gz

certik:
I see. Fixed.

ahmadia:
+ 1

ahmadia:
err, +1

certik:
Thanks for the review. You know that you can modify comments. ;)

robertmaynard:
Have you looked at using ```--depth 1``` with git to make checkouts faster as that does a shallow clone ( no history )?

Just an FYI: Certain things that are possible with makefiles are impossible with ninja currently, so switching from make to ninja as the CMake generator could cause builds to fail. It is fairly rare, but is usually caused by incomplete listing of outputs from ```add_custom_commands```, ```add_custom_commands``` with the ```POST_BUILD ```  options, or incomplete consumption of outputs from ```custom_commands```

certik:
Also ninja doesn't work for Fortran, so I can't use it myself or my work anyway.

@robertmaynard, I got the idea from @thewtex who suggested [here](https://plus.google.com/u/0/111200425937199925962/posts/1DywGBxDXTt) that "the Ninja generator is highly recommended!﻿", but the Fortran support needs to be added first.

ahmadia:
> Have you looked at using --depth 1 with git to make checkouts faster as that does a shallow clone ( no history )?

This was one of the very first suggestions I made to Dag when I joined the team.  Because we are grabbing commits and not refs, we can't actually use `--depth 1`, since there is no way to request a remote commit from a repository, only a ref.

robertmaynard:
@certik For well behaved CMake projects that only use C/C++ ninja is a good choice.  Brad King has proposed adding Fortran support to ninja, but that is still in progress.

PR #755, opened 2015-05-08
added scored mesh tools and zoltan partitioner dep.

cekees:


pkgs/scorec.yaml
```diff
@@ -0,0 +1,36 @@
+extends: [cmake_package]
+dependencies:
+  build: [mpi, parmetis, zoltan, simmetrix]
+
+sources:
+- key: git:ccd43aaab35fbc471f24f20860025985595a3b3a
+  url: git@bitbucket.org:ibaned/scorec.git
```
certik: Shouldn't this rather be https instead of git?

pkgs/zoltan.yaml
```diff
@@ -0,0 +1,32 @@
+extends: [autotools_package]
+dependencies:
+  build: [mpi, {{build_with}}]
+
+sources:
+- key: tar.gz:lpoumvepxhdtwis3xtz5ebwfldbrrszj
+  url: http://www.cs.sandia.gov/~kddevin/Zoltan_Distributions/zoltan_distrib_v3.8.tar.gz
```
certik: Isn't Zoltan part of Trilinos?

certik:
Otherwise +1.

cekees:
@ibaned, can I just point at https://github.com/SCOREC/core now? Is the idea to point at core-sim if one has a license to the simmetrix tools and access to that repo, otherwise point at core, but otherwise  the two  repos supply the same interface?

ibaned:
you can point to https://github.com/SCOREC/core or https://github.com/SCOREC/core-sim, but definitely not to bitbucket anymore (active development / support is on Github now). "core" is completely public and "core-sim" is restricted to our collaborators. there is a difference in what is available and how it should be compiled/linked. here are some calls for example that would not be executable with "core" alone:

https://github.com/erdc-cm/proteus-scorec/blob/master/src/MeshAdaptPUMI/MeshAdaptPUMI.cpp#L15

https://github.com/erdc-cm/proteus-scorec/blob/master/src/MeshAdaptPUMI/MeshAdaptPUMI.cpp#L40

other than that, the overwhelming majority of the code ought to compile fine with "core", though you would not have snapping to curved geometry during adaptation since that is done through Simmetrix CAD interfaces.

ibaned:
Zoltan is part of Trilinos but can be built separately. It is one of a few projects that used to be independent but then were integrated to build within Trilinos, but it does not use other Trilinos packages. We typically build it separately and that cs.sandia.gov will get you a version that supports the SCOREC tools effectively.

cekees:
@certik I'd like to go ahead and  merge this. The issue of whether to build Zoltan inside trilinos is open. I like what we've done with the petsc parameterization, which supports building packages both with petsc's build system and outside  of it. I suspect we should try to do the same with trilinos.

pkgs/zoltan.yaml
```diff
@@ -0,0 +1,32 @@
...
+    bash: |
+      mkdir -p _build
+      cd _build
+
+  - name: setEnv
+    after: setup_builddir
+    handler: bash
+    bash: |
+      export MPI_DIR=$(dirname $(dirname $MPICC) )
+      export LD_LIBRARY_PATH=${MPI_DIR}/lib
```
certik: This is very hackish to set `LD_LIBRARY_PATH`, even if during the build only. Usually there is a better solution. But that's fine for now to get something that works.

certik:
Thanks @ibaned for the clarification.

That's fine for now, merging.

PR #760, opened 2015-05-08
added pygrib and dependencies

cekees:


pkgs/pygrib.yaml
```diff
@@ -0,0 +1,21 @@
+extends: [distutils_package]
+
+dependencies:
+  build: [python,jasper,grib_api,pyproj,numpy]
+  run: [python,pyproj]
```
certik: Add spaces after comma

pkgs/pygrib.yaml
```diff
@@ -0,0 +1,21 @@
...
+
+sources:
+- key: tar.gz:u4hhkrnmmbciyv2qjzajx4ne3em2l5mb
+  url: https://pygrib.googlecode.com/files/pygrib-1.9.8.tar.gz
```
certik: Google code is going down soon. We need a better upstream url.

pkgs/pyproj.yaml
```diff
@@ -0,0 +1,8 @@
+extends: [distutils_package]
+dependencies:
+  build: [python]
+  run: [python]
+
+sources:
+- key: tar.gz:sexnq3c6ib7qshiphudwlffpj62wtyt4
+  url: https://pyproj.googlecode.com/files/pyproj-1.9.3.tar.gz
```
certik: Google code is going down soon. We need a better upstream url.

PR #761, opened 2015-05-08
pexpect 3.3 from github zip

cekees:


jcftang:
+1

pkgs/pexpect.yaml
```diff
@@ -0,0 +1,9 @@
+extends: [setuptools_package]
+
+dependencies:
+  build: []
+  run: []
```
certik: This `dependencies` section can be removed.

certik:
Otherwise +1.

PR #762, opened 2015-05-08
openjpeg 1.5.1 from dev site, used by grib_api

cekees:
Note: I couldn't get pillow to use this as its jpeg2000 lib

### DIFF
--- a/pkgs/openjpeg.yaml
+++ b/pkgs/openjpeg.yaml
@@ -1,0 +1,24 @@
+extends: [cmake_package]
+
+# version 1.5 of openjpeg
+# later versions should go in openjpeg2
+
+dependencies:
+  build: []
+
+defaults:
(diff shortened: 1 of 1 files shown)


pkgs/openjpeg.yaml
```diff
@@ -0,0 +1,24 @@
...
+
+dependencies:
+  build: []
+
+defaults:
+  relocatable: false
+
+sources:
+- key: tar.gz:nt5tbf5bzduvkajrwf3ghwxze72cob54
+  url: https://openjpeg.googlecode.com/files/openjpeg-1.5.1.tar.gz
```
certik: Google code is going down soon. We need a better upstream url.

cekees:
pillow will build  with openjpeg when #764 goes in.

PR #764, opened 2015-05-10
added options for <#> versions of png(zlib), freetype, and jpeg2K

cekees:
added the bits that point to the hashstack versions of  the dependencies and enforces  the related support in PIL.

### DIFF
--- a/pkgs/pillow.yaml
+++ b/pkgs/pillow.yaml
@@ -12,7 +12,4 @@ build_stages:
   before: install
   handler: bash
   bash: |
-       export SETUPTOOLS_PACKAGE_EXTRA_OPTIONS=build_ext \
-       --enable-zlib --enable-freetype --enable-jpeg2000 \
-       --include-dirs="${ZLIB_DIR}/include:${PNG_DIR}/include:${FREETYPE_DIR}/include:${OPENJPEG_DIR}/include" \
-       --library-dirs="${ZLIB_DIR}/lib:${PNG_DIR}/lib:${FREETYPE_DIR}/lib:${OPENJPEG_DIR}/lib"
+       export SETUPTOOLS_PACKAGE_EXTRA_OPTIONS="build_ext --enable-zlib --enable-freetype --enable-jpeg2000 --include-dirs=${ZLIB_DIR}/include:${PNG_DIR}/include:${FREETYPE_DIR}/include:${OPENJPEG_DIR}/include --library-dirs=${ZLIB_DIR}/lib:${PNG_DIR}/lib:${FREETYPE_DIR}/lib:${OPENJPEG_DIR}/lib"
--- a/pkgs/pillow.yaml
+++ b/pkgs/pillow.yaml
@@ -1,6 +1,6 @@
 extends: [setuptools_package]
 dependencies:
-  build: [zlib,png,freetype,openjpeg]
+  build: [zlib, png, freetype, openjpeg]
 
 sources:
 - key: tar.gz:q5qmcgfaefplufr7o6bbcdt67tn3cx4k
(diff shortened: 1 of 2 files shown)


pkgs/pillow.yaml
```diff
@@ -1,7 +1,21 @@
 extends: [setuptools_package]
 dependencies:
-  build: [png]
+  build: [zlib,png,freetype,openjpeg]
```
certik: Add spaces after commas

pkgs/pillow.yaml
```diff
@@ -1,7 +1,21 @@
...
 sources:
 - key: tar.gz:q5qmcgfaefplufr7o6bbcdt67tn3cx4k
   url: https://pypi.python.org/packages/source/P/Pillow/Pillow-2.8.1.tar.gz
+
+build_stages:
+- name: install
+  mode: replace
+  handler: bash
+  bash: |
+       ${PYTHON} -c 'import setuptools; __file__="setup.py"; exec(open(__file__).read())' \
```
certik: This looks really messed up --- this hack should only happen in the `setuptools_package.yaml` I would think. I am ok with this as a temporary workaround to make it do what you need, but as a long term solution I would like to figure out a way to perhaps add some parameters to `setuptools_package.yaml` that you just set here. Doesn't just adding `libflags` into `extends` fix this issue?

certik: @cekees  How about something like this:
```diff
diff --git a/base/setuptools_package.yaml b/base/setuptools_package.yaml
index 9818599..5f875fd 100644
--- a/base/setuptools_package.yaml
+++ b/base/setuptools_package.yaml
@@ -22,6 +22,7 @@ build_stages:
   handler: bash
   bash: |
     ${PYTHON} -c 'import setuptools; __file__="setup.py"; exec(open(__file__).read())' \
+       ${SETUPTOOLS_PACKAGE_EXTRA_OPTIONS} \
        install \
        --prefix=. --root=${ARTIFACT} \
        --single-version-externally-managed
diff --git a/pkgs/pillow.yaml b/pkgs/pillow.yaml
index fff105c..c95709f 100644
--- a/pkgs/pillow.yaml
+++ b/pkgs/pillow.yaml
@@ -7,15 +7,11 @@ sources:
   url: https://pypi.python.org/packages/source/P/Pillow/Pillow-2.8.1.tar.gz
 
 build_stages:
-- name: install
-  mode: replace
+- name: install2
+  before: install
   handler: bash
   bash: |
-       ${PYTHON} -c 'import setuptools; __file__="setup.py"; exec(open(__file__).read())' \
-       build_ext \
+       SETUPTOOLS_PACKAGE_EXTRA_OPTIONS = build_ext \
        --enable-zlib --enable-freetype --enable-jpeg2000 \
        --include-dirs="${ZLIB_DIR}/include:${PNG_DIR}/include:${FREETYPE_DIR}/include:${OPENJPEG_DIR}/include" \
-       --library-dirs="${ZLIB_DIR}/lib:${PNG_DIR}/lib:${FREETYPE_DIR}/lib:${OPENJPEG_DIR}/lib" \
-       install \
-       --prefix=. --root=${ARTIFACT} \
-       --single-version-externally-managed
+       --library-dirs="${ZLIB_DIR}/lib:${PNG_DIR}/lib:${FREETYPE_DIR}/lib:${OPENJPEG_DIR}/lib"
```

cekees:
ok, this works on my linux machine.

cekees:
@dagss this package is a good case for  the "?" syntax. PIL has a slew of optional dependencies that range in complexity from zlib to  qt4.

certik:
I am testing this PR on my linux.

certik:
Ok, it works on my linux machine as well. Merging.

dagss:
Thanks @cekees looks like it!

PR #768, opened 2015-05-14
add -R switch to ode python bindings

cekees:
This eliminates the need for setting LD_LIBRARY_PATH when running ode from python.


certik:
Looks good.

PR #773, opened 2015-05-15
git: allow to set NO_APPLE_COMMON_CRYPTO=1

certik:
This fixes a build failure on OS X 10.10. Unfortunately, this disabled https
access, so it is only enabled if 'disable_crypto' parameter is 'true'.

### DIFF
--- a/pkgs/git.yaml
+++ b/pkgs/git.yaml
@@ -10,6 +10,7 @@ sources:
 defaults:
   # /bin/git contains hard-coded path
   relocatable: false
+  disable_crypto: false
 
 build_stages:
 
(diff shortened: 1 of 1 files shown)


certik:
I should note that git builds fine with the native clang, but fails with gcc. So a profile with clang doesn't need any parameter, but a profile with gcc requires `disable_crypto: true`.

jcftang:
+1

certik:
Thanks. By default it doesn't do anything, and it optionally allows to make git build on OS X with gcc, albeit not ideally. We can improve upon this later.

PR #774, opened 2015-05-15
Use Clang's linker to make AVX instructions work

### LINKED ISSUE #748: OS X 10.10 using '-march=native' in works in clang but breaks in gcc
More background e.g. in https://github.com/Homebrew/homebrew/issues/14204.
The way it breaks in (our compiled) gcc is as follows:
```
[csympy] Building CXX object src/CMakeFiles/csympy.dir/symbol.cpp.o
[csympy] /var/tmp//ccc4Ii22.s:829:no such instruction: `vzeroupper'
[csympy] /var/tmp//ccc4Ii22.s:847:no such instruction: `vzeroupper'
[csympy] /var/tmp//ccc4Ii22.s:855:no such instruction: `vzeroupper'
...
```
The linker (which is Apple linker) complains that it doesn't understand the AVX instructions, emitted (correctly) by gcc's `-march=native`, since my computer supports AVX. A quick workaround is to do
```diff
diff --git a/pkgs/csympy.yaml b/pkgs/csympy.yaml
index 5caa557..32bf39d 100644
--- a/pkgs/csympy.yaml
+++ b/pkgs/csympy.yaml
@@ -12,6 +12,12 @@ defaults:
   relocatable: false
 
 build_stages:
+- name: cxx_flags
+  before: configure
+  handler: bash
+  bash: |
+    export CXXFLAGS="-mno-avx"
+
 - name: configure
   extra: ['-D WITH_PYTHON:BOOL=ON',
           '-D PYTHON_INSTALL_PATH:PATH=$ARTIFACT/{{python_site_packages_rel}}']
```
But this is not a solution, because you want to be able to use AVX instructions if your processor supports it, and gcc generates them.

The fix is to either make gcc generate AVX instructions so that the Apple's linker can understand, or we need to build our own linker.

certik:
Fixes #748

However, I am not happy that we need to patch every package that uses our gcc. Rather, this patch should probably go into some base template, so that all packages use it.

### DIFF
--- a/pkgs/csympy.yaml
+++ b/pkgs/csympy.yaml
@@ -12,6 +12,12 @@ defaults:
   relocatable: false
 
 build_stages:
+- name: cxx_flags
+  before: configure
+  handler: bash
+  bash: |
+    export CXXFLAGS="-Wa,-q"
+
 - name: configure
   extra: ['-D WITH_PYTHON:BOOL=ON',
           '-D PYTHON_INSTALL_PATH:PATH=$ARTIFACT/{{python_site_packages_rel}}']


certik:
@ahmadia, @cekees what is the best way to handle this? The `-Wa,-q` flag needs to be added to every package that is compiled with gcc **and** with `-march=native` on OS X. Without it, the linking fails with missing AVX instructions (the flag forces gcc to use the correct OS X linker which understands these instructions):
```
[bddc] /var/tmp//ccYZ86gg.s:46:no such instruction: `vmovsd (%r14,%rax,8), %xmm0'
[bddc] /var/tmp//ccYZ86gg.s:57:no such instruction: `vmovsd (%r11,%r9,8), %xmm1'
[bddc] /var/tmp//ccYZ86gg.s:61:no such instruction: `vcomisd %xmm0, %xmm1'
[bddc] /var/tmp//ccYZ86gg.s:69:no such instruction: `vmovsd (%r11,%r15,8), %xmm2'
[bddc] /var/tmp//ccYZ86gg.s:74:no such instruction: `vcomisd %xmm2, %xmm0'
[bddc] /var/tmp//ccYZ86gg.s:87:no such instruction: `vmovsd (%r9), %xmm3'
...
```
So only if gcc and OS X is used, the flag should be set in all `CXXFLAGS`, `CFLAGS` and `FCFLAGS`. The problem only manifests if the `-march=native` option is used, but I think it doesn't hurt to use the flag always.

One way is to set this in the base template yaml for OS X, and then change all Hashstack to only keep appending to `CXXFLAGS`, `CFLAGS` and `FCFLAGS`, instead of replacing what was there (the current approach).

cekees:
I'm not familiar with the `-Wa,-q` option. If it doesn't hurt anything, maybe we should have an os x profile that adds it and `-march=native` to the prologue.

certik:
Right. Except that we still need to update all Hashstack to keep appending to `CXXFLAGS`, `CFLAGS` and `FFLAGS` as I mentioned above.

PR #771, opened 2015-05-15
gcc: install fixed dispatch/object.h header file

certik:
Fixes #742.

I tested this, it is ready for review.

PR #776, opened 2015-05-15
Add hd_base example profile for OS X

certik:
@cekees, @ahmadia let me know what you think.

This profile requires #771 to be merged (well, strictly speaking it will build even without it, but then some other packages that you try to build with gcc will fail, so it's better to have it in, then you do not need to recompile once it is merged). I used the latest master (4637543).

### DIFF
--- a/examples/hd_base.Darwin.yaml
+++ b/examples/hd_base.Darwin.yaml
@@ -1,0 +1,41 @@
+# This profile builds with the default clang compiler on Mac OS X 10.10
+# Yosemite. In particular it builds gcc (g++, gcc, gfortran). This profile can
+# then be used as a base for building the rest of Hashstack.
+
+extends:
+ - file: config.yaml
+
+parameters:
+  platform: Darwin
+  fortran: false
+  PATH: /usr/bin:/bin:/usr/sbin:/sbin
+  PROLOGUE: |
+    export MACOSX_DEPLOYMENT_TARGET=$(sw_vers -productVersion | sed -E "s/([0-9]+\.[0-9]+).*/\1/")
+
+packages:
+
+  launcher:
+  blas:
+    use: host-osx-framework-accelerate
+  lapack:
+    use: host-osx-framework-accelerate
+  mpi:
+    use: mpich
+  python:
+    link: shared
+  swig:
+    build_with: |
+      perl
+  ipython:
+  nose:
+  numpy:
+  sphinx:
+  sympy:
+  gmp:
+  matplotlib:
+  doxygen:
+  breathe:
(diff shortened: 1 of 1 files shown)


cekees:
It  looks  good  to me. Why is `fortran: false` required?

certik:
Because OS X does not have a Fortran compiler. See the comment in the file. This profile is a base profile that you build with the OS X native tools (in fact, just a pristine install of OS X 10.10 plus the default command line tools installation, triggered by typing "gcc" into the terminal the first time, is needed).

Once you build this profile, then you have decent compilers and can compile the rest of Hashstack (with Fortran).

cekees:
I realize that. My question is why you need to set the parameter in a base profile. I guess you have packages in this base profile that may fail to build unless they're explicitly told not to build with fortran support.  Seems to me that one would generally wait to build those in the downstream profile based on the gcc compilers built here.  It's not a problem with your PR. I just  want to make sure I'm clear on what is happening.

certik:
Ah I see. Yes, it's probably not necessary, but it's nice to have it there as a remainder that this profile, if used as directed in the comments, will not build the whole Hashstack, only a minimal profile.

examples/hd_base.Darwin.yaml
```diff
@@ -0,0 +1,41 @@
...
+
+packages:
+
+  launcher:
+  blas:
+    use: host-osx-framework-accelerate
+  lapack:
+    use: host-osx-framework-accelerate
+  mpi:
+    use: mpich
```
ahmadia: You won't get the Fortran compiler wrappers in this profile if you build MPI this way.

ahmadia:
I'm not sure what the purpose of this profile is.  If it's just a builder, I would imagine that it only contains things like compilers, automake, etc...  If it's something else, I'm not really sure what I'd use it for.

examples/hd_base.Darwin.yaml
certik: You are right, the mpi will build without Fortran. To build it with Fortran, one needs to use #779.

certik:
It's a builder, so it has gcc, cmake, python, git and few other things that you might need to build packages, either as part of Hashstack, or manually. You are right about the MPI, that it doesn't have Fortran support. 

I personally use this profile as a base profile for #779.

certik:
We can also later extend it to include `hashdist` itself, very useful for installations that do not have lzma or other dependencies installed.

Anyway, any objections against merging this?

cekees:
No objections.

certik:
I am merging this. We can improve upon it later.

Issue #368, opened 2014-07-16
Building CMake fails on Darwin


johannesring:
The following profile

```
extends:
- file: osx.yaml

packages:
  cmake:
```

fails with:

```
...
[cmake] -- The C compiler identification is Clang 5.1.0
[cmake] -- The CXX compiler identification is Clang 5.1.0
[cmake] CMake Error at Modules/Platform/Darwin.cmake:211 (message):
[cmake]   CMAKE_OSX_DEPLOYMENT_TARGET is '10.9' but CMAKE_OSX_SYSROOT:
[cmake] 
[cmake]    ""
[cmake] 
[cmake]   is not set to a MacOSX SDK with a recognized version.  Either set
[cmake]   CMAKE_OSX_SYSROOT to a valid SDK or set CMAKE_OSX_DEPLOYMENT_TARGET to
[cmake]   empty.
[cmake] Call Stack (most recent call first):
[cmake]   Modules/CMakeSystemSpecificInformation.cmake:36 (include)
[cmake]   CMakeLists.txt:14 (project)
[cmake] 
[cmake] 
[cmake] -- Configuring incomplete, errors occurred!
[cmake] See also "/Users/buildbot/.hashdist/tmp/cmake-ckvi5ak55y2k-1/CMakeFiles/CMakeOutput.log".
[cmake] ---------------------------------------------
[cmake] Error when bootstrapping CMake:
[cmake] Problem while running initial CMake
[cmake] ---------------------------------------------
[cmake] make: *** No targets specified and no makefile found.  Stop.
[cmake ERROR] Command '[u'/bin/bash', '_hashdist/build.sh']' returned non-zero exit status 2
[cmake ERROR] command failed (code=2); raising
```

certik:
@robertmaynard I tried setting `-DCMAKE_OSX_DEPLOYMENT_TARGET:STRING=""`, but cmake still picks up the environment variable `MACOSX_DEPLOYMENT_TARGET` (10.10 on my machine) and sets `CMAKE_OSX_DEPLOYMENT_TARGET` to 10.10., and then it fails with:
```
[blas] CMake Error at /Users/certik/.hashdist/bld/cmake/awipsecciam6/share/cmake-3.2/Modules/Platform/Darwin.cmake:76 (message):
[blas]   CMAKE_OSX_DEPLOYMENT_TARGET is '10.10' but CMAKE_OSX_SYSROOT:
[blas] 
[blas]    ""
[blas] 
[...]

PR #780, opened 2015-05-18
Unset MACOSX_DEPLOYMENT_TARGET for all cmake

certik:
This is needed, otherwise cmake fails to compile packages.

### DIFF
--- a/base/cmake_package.yaml
+++ b/base/cmake_package.yaml
@@ -12,6 +12,14 @@ build_stages:
       mkdir -p _build
       cd _build
 
+  - when: platform == 'Darwin'
+    name: fix_deployment_target
+    after: prologue
+    before: configure
+    handler: bash
+    bash: |
+      unset MACOSX_DEPLOYMENT_TARGET
+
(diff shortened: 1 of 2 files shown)


ahmadia:
This is certainly "new".  This is sometimes needed to make Python packages build correctly, and for enforcing deployment targets.

certik:
There might be some way to convince CMake to work with MACOSX_DEPLOYMENT_TARGET. I found, as you said, that Python (and perhaps more packages) require MACOSX_DEPLOYMENT_TARGET to be properly set so I set it in PROLOGUE. However, cmake always fails to build other packages, if this is set --- so one solution is to unset MACOSX_DEPLOYMENT_TARGET just for packages build using cmake, the other solution is to figure out how to make cmake work with it (I think that would be preferable, probably). This PR implements the former.

johannesring:
See also issue #368.

certik:
See also #/902. It looks like on El Capitan at least, things are working as expected.

PR #782, opened 2015-05-19
boost: run patchelf on the libraries, not on the symlinks

### LINKED ISSUE #781: ImportError: libbz2.so.1.0: cannot open shared object file: No such file or directory
I have successfully built FEniCS with Hashdist on a cluster in Sweden, but when I try to import dolfin in Python I get the following error:

    $ python -c "import dolfin"
    Traceback (most recent call last):
      File "<string>", line 1, in <module>
      File "/home/johannr/src/hashstack/fenics-1.5.0.alarik.lunarc.gnu/lib/python2.7/site-packages/dolfin/__init__.py", line 16, in <module>
        from . import cpp
      File "/home/johannr/src/hashstack/fenics-1.5.0.alarik.lunarc.gnu/lib/python2.7/site-packages/dolfin/cpp/__init__.py", line 42, in <module>
        exec("from . import %s" % module_name)
      File "<string>", line 1, in <module>
      File "/home/johannr/src/hashstack/fenics-1.5.0.alarik.lunarc.gnu/lib/python2.7/site-packages/dolfin/cpp/common.py", line 32, in <module>
        _common = swig_import_helper()
      File "/home/johannr/src/hashstack/fenics-1.5.0.alarik.lunarc.gnu/lib/python2.7/site-packages/dolfin/cpp/common.py", line 28, in swig_import_helper
        _mod = imp.load_module('_common', fp, pathname, description)
    ImportError: libbz2.so.1.0: cannot open shared object file: No such file or directory

It turns out that the rpath has not been set correctly in the Boost iostreams library:

    $ ldd /alarik/nobackup/i_p/johannr/.hashdist/bld/boost/y7he5zxfcmvv/lib/libboost_iostreams.so.1.55.0 | grep bz2
    	libbz2.so.1.0 => not found

However, it is set correctly in the symbolic link:

    $ ldd /alarik/nobackup/i_p/johannr/.hashdist/bld/boost/y7he5zxfcmvv/lib/libboost_iostreams.so | grep bz2
    	libbz2.so.1.0 => /alarik/nobackup/i_p/johannr/.hashdist/bld/boost/y7he5zxfcmvv/lib/../../../bzip2/ydfav75zjrep/lib/libbz2.so.1.0 (0x00007fbc8f200000)

But this is not a symbolic link as it should be:

    $ ls -l /alarik/nobackup/i_p/johannr/.hashdist/bld/boost/y7he5zxfcmvv/lib/libboost_iostreams.so*
    -r-xr-xr-x 1 johannr lu-jhsimu 160312 May 19 12:20 /alarik/nobackup/i_p/johannr/.hashdist/bld/boost/y7he5zxfcmvv/lib/libboost_iostreams.so
    -r-xr-xr-x 1 johannr lu-jhsimu 160176 May 19 12:20 /alarik/nobackup/i_p/johannr/.hashdist/bld/boost/y7he5zxfcmvv/lib/libboost_iostreams.so.1.55.0

The problem is in `boost.yaml`, where we run patchelf on the symbolic links instead of the actual library files. This turns them into copies instead of symbolic links. I will make a PR that fixes this problem.

johannesring:
This will fix #781.

### DIFF
--- a/pkgs/boost/boost.yaml
+++ b/pkgs/boost/boost.yaml
@@ -91,7 +91,7 @@ build_stages:
   after: bjam
   handler: bash
   bash: |
-    for lib in ${ARTIFACT}/lib/*.so; do
+    for lib in ${ARTIFACT}/lib/*.so.*; do
         ${PATCHELF} --set-rpath ${ARTIFACT}/lib:${BZIP2_DIR}/lib:${ZLIB_DIR}/lib ${lib}
     done
 


certik:
+1. We should go over all patchelf uses in Hashstack and make sure this bug is not there.

johannesring:
Thanks, I'm merging this now (the timeouts for travis are not related).

We actually had the same problem with HYPRE recently, which was fixed in 5815793cb2048ec6eb503d9f95ae2d0bc44513af.

PR #789, opened 2015-05-22
Add Armadillo package

chvillanuevap:
Armadillo does not support out-of-source builds.

### DIFF
--- a/pkgs/armadillo.yaml
+++ b/pkgs/armadillo.yaml
@@ -1,0 +1,18 @@
+extends: [cmake_package]
+
+defaults:
+  relocatable: false
+
+sources:
+- key: tar.gz:mmkq6ev7pr6ftdvflnzhdwm7zy2j744g
+  url: http://sourceforge.net/projects/arma/files/armadillo-5.100.2.tar.gz
+
+build_stages:
+  - name: setup_builddir
+    after: prologue
+    mode: replace
+    handler: bash
+    bash: |
+
+  - name: configure
+    build_in_source: true


certik:
I tested it on my RHEL6, works great. Thanks! Merging.

chvillanuevap:
Just for future reference, you can build Armadillo with out-of-source build settings. I tested this successfully in MAC OS X Yosemite. However, in their README.txt file, they mention that out-of-source builds are not fully supported, and recommend using an in-source one.

johannesring:
This is a good start but I think it is a little bit incomplete.

We used to depend on Armadillo for DOLFIN and I remember it was often a problem that it wasn't linked correctly with BLAS/LAPACK. We need to be careful about this.

Notice this part of the [build.log](https://gist.github.com/johannesring/2279f8992fc0685449a8):

    [armadillo] -- Found BLAS: /usr/lib/libblas.so
    [armadillo] -- Found LAPACK: /usr/lib/liblapack.so
    [armadillo] --      MKL_FOUND = NO
    [armadillo] --   ACMLMP_FOUND = NO
    [armadillo] --     ACML_FOUND = NO
    [armadillo] -- OpenBLAS_FOUND = NO
    [armadillo] --    ATLAS_FOUND = NO
    [armadillo] --     BLAS_FOUND = YES
    [armadillo] --   LAPACK_FOUND = YES
    [armadillo] -- Found ARPACK: /usr/lib/libarpack.so
    [armadillo] -- ARPACK_FOUND = YES
    [armadillo] -- Found SuperLU: /usr/lib/x86_64-linux-gnu/libsuperlu.so
    [armadillo] -- SuperLU_FOUND = YES
    [armadillo] -- SuperLU_INCLUDE_DIR = /usr/include/superlu
    [armadillo] -- 
    [armadillo] -- *** Armadillo wrapper library will use the following libraries:
    [armadillo] -- *** ARMA_LIBS = /usr/lib/libblas.so;/usr/lib/liblapack.so;/usr/lib/libarpack.so;/usr/lib/x86_64-linux-gnu/libsuperlu.so

We can see that it picks up BLAS, LAPACK, ARPACK and SuperLU from the system. These libraries should be added to the build dependencies to make sure Armadillo use the ones built by HashDist.

We do not use Armadillo anymore for DOLFIN, so this is not important for me. I just wanted to let you know. I will be glad to help out of course.

chvillanuevap:
That's a great suggestion actually! In my group, we compile separate executables with LAPACK, ACML, and MKL.

I have a few questions actually:

1. Can we override dependencies in the profile? For example, if the package Armadillo has the following build dependencies:
```
dependencies:
  build: [blas, lapack, arpack, superlu]
```
can we have the following in the profile to use a different blas library?
```
packages:
  armadillo:
    dependencies:
      build: [openblas, arpack, superlu]
```
or do we need to replace it indirectly through blas and lapack?
```
packages:
  blas:
    use: openblas
  lapack:
    use: openblas
  armadillo:
```

2. What is the difference between build and run in the dependencies node? I have seen both been used in HashStack.
```
dependencies:
  build: [blas, lapack, arpack, superlu]
  run: [blas, lapack, arpack, superlu]
```

3. This last question is unrelated to this PR, but I was wondering why dolfin dropped Armadillo? We are starting development of a new research code in my group, and we debating between Armadillo and Eigen3. Is there a mailing group conversation that I can see where this was discussed?

certik:
Yes, you can override the dependencies in the profile. First of all, for blas/lapack, one can use:
```
packages:
  blas:
    use: openblas
  lapack:
    use: openblas
  armadillo:
```
Usually you want to use the same lapack/blas for everything in the profile. If this is not the case for you, let us know --- i.e. do you need to build Armadillo with a different lapack implementation than let's say numpy?

For other dependencies, you can follow the `python` package, which allows you to add dependencies in the profile.

chvillanuevap:
No, a single lapack is what I need. I see what you mean with the python package. Is there a reason why {{build_with}} isn't used in all packages by default? For example, I noticed Trilinos does not have it, and there are a lot more libraries that I could use with it.

certik:
That's a good point, and I don't have a good answer to that. @dagss, @cekees, @ahmadia, what do you think?

I know sometimes the build script requires some modification to add a new package, for example you have to point cmake to it, but in some other cases, no modification is needed, since the standard `cmake_package` template adds the necessary include flags automatically, so cmake picks it up without a modification.

dagss:
@chvillanuevap , work is underway now to change this aspect of Hashdist. So right now the best you can do is that `use:` statement, but in some weeks you should be able to do:

```
dependencies:
    build: [blas, lapack, arpack, superlu?]  # assuming, say, superlu is optional
    run: [blas, lapack, arpack, superlu?]
```

and then in the profile

```
packages:
  armadillo:
    blas: openblas
    lapack: openblas
  superlu:  # since this is in profile, it will be auto-included in armadillo too
  ..
```

johannesring:
@chvillanuevap - Here is the discussion on the FEniCS mailing list: [Switching from Armadillo to Eigen](http://fenicsproject.org/pipermail/fenics/2013-September/000569.html)

johannesring:
The problem with using `{{build_with}}` is that you currently have to write a Python file to check if an optional package is in `{{build_with}}`, like we have done for PETSc and DOLFIN. The work done by @dagss will hopefully make this easier.

PR #790, opened 2015-05-23
Refactor out-of-source build with CMake

chvillanuevap:
This Armadillo package:

```
build_stages:

  - name: configure
    build_in_source: true
```

produces the following build script:

```
set -e
export HDIST_IN_BUILD=yes
export MACOSX_DEPLOYMENT_TARGET=$(sw_vers -productVersion | sed -E "s/([0-9]+\.[0-9]+).*/\1/")

(
export CPPFLAGS="-I${CMAKE_DIR}/include"
export LDFLAGS="-L${CMAKE_DIR}/lib "
${CMAKE} -DCMAKE_INSTALL_PREFIX:PATH="${ARTIFACT}" \
  -DCMAKE_POSITION_INDEPENDENT_CODE:BOOL=ON \
  -DCMAKE_INSTALL_RPATH:STRING="${ARTIFACT}/lib" \
  -DCMAKE_INSTALL_RPATH_USE_LINK_PATH:BOOL=ON \
  -DCMAKE_BUILD_TYPE:STRING=Release \
  -DCMAKE_PREFIX_PATH="${CMAKE_DIR}" \
  .
)
${CMAKE} --build . -- -j ${HASHDIST_CPU_COUNT}
make install
```

while this version of the package:

```
build_stages:

  - name: configure
    build_in_source: false
```

generates:

```
set -e
export HDIST_IN_BUILD=yes
export MACOSX_DEPLOYMENT_TARGET=$(sw_vers -productVersion | sed -E "s/([0-9]+\.[0-9]+).*/\1/")
mkdir -p _build && cd _build
(
export CPPFLAGS="-I${CMAKE_DIR}/include"
export LDFLAGS="-L${CMAKE_DIR}/lib "
${CMAKE} -DCMAKE_INSTALL_PREFIX:PATH="${ARTIFACT}" \
  -DCMAKE_POSITION_INDEPENDENT_CODE:BOOL=ON \
  -DCMAKE_INSTALL_RPATH:STRING="${ARTIFACT}/lib" \
  -DCMAKE_INSTALL_RPATH_USE_LINK_PATH:BOOL=ON \
  -DCMAKE_BUILD_TYPE:STRING=Release \
  -DCMAKE_PREFIX_PATH="${CMAKE_DIR}" \
  ..
)
${CMAKE} --build . -- -j ${HASHDIST_CPU_COUNT}
make install
```

### DIFF
--- a/base/cmake_package.py
+++ b/base/cmake_package.py
@@ -51,9 +51,14 @@ def configure(ctx, stage_args):
     if 'extra' in stage_args:
         conf_lines.append(' '.join('%s' % arg for arg in stage_args['extra']))
 
+    #set in-source or out-of-source build
+    setup_lines = []
+    mkdirbuild = 'mkdir -p _build && cd _build'
     builddir = stage_args.get('builddir', '..')
     if stage_args.get('build_in_source', False):
+	mkdirbuild = ''
         builddir = '.'
+    setup_lines.append(mkdirbuild)
     conf_lines.append(builddir)
 
     for i in range(len(conf_lines) - 1):
@@ -65,4 +70,4 @@ def configure(ctx, stage_args):
         env_lines.append('export CPPFLAGS="%s"' % ' '.join(CPPFLAGS))
         env_lines.append('export LDFLAGS="%s"' % ' '.join(LDFLAGS))
 
-    return ['('] + env_lines + conf_lines + [')']
+    return setup_lines + ['('] + env_lines + conf_lines + [')']
--- a/base/cmake_package.yaml
+++ b/base/cmake_package.yaml
@@ -5,15 +5,8 @@ dependencies:
 
 build_stages:
 
-  - name: setup_builddir
-    after: prologue
-    handler: bash
-    bash: |
-      mkdir -p _build
-      cd _build
-
   - name: configure
-    after: setup_builddir
+    after: prologue
     debug: {{debug}}
 
   - name: make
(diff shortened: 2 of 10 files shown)


chvillanuevap:
Upon further review, I do not know how this could work with ParMETIS. The issue is that ParMETIS needs two separate build directories: `_build_metis` and `build_parmetis`.

We could create another variable like `build_in_source` called `build_dir`:

```
build_stages:

  - name: configure
    build_in_source: false
    build_dir: _build_metis, _build_parmetis
```

What do you think?

certik:
Parmetic is not build using the cmake template (it calls cmake directly):
```
extends: [base_package]
```
so you do not need to worry about it in this PR.

This PR looks good to me. We can now remove the part
```diff
-  - name: setup_builddir
-    after: prologue
-    handler: bash
-    bash: |
-      mkdir -p _build
-      cd _build
```
from all packages. Before we merge this, I would like to test it on my larger stack, to make sure things still work.

chvillanuevap:
I see what you mean. I saw CMake being used in Parmetis and missed that it depends on the base_package, not the cmake_package. I already modified the other packages. I will commit them at some point tomorrow.

chvillanuevap:
It seems dolfin sets up its CMake file independent of the `cmake_package` configuration file. I duplicated the code from `cmake_package.py` to `dolfin.py`.

chvillanuevap:
I'm going to close this PR, create a branch, and resubmit it. I should not have committed this directly to my master branch.

I want to do this in order to submit a few new packages I have implemented.

certik:
Ok. Please ping me when it is ready. This is a good change.

PR #799, opened 2015-06-08
Update pkg-config.yaml

agondolo:
Old link doesn't work anymore. New source added.


agondolo:
Another option: http://files.etoilebsd.net/pkgconf/pkgconf-0.9.6.tar.bz2

pkgs/pkg-config/pkg-config.yaml
```diff
@@ -10,7 +10,7 @@ defaults:
 
 sources:
 - key: tar.bz2:yqfhovb74xdslg4bh5yn3lzctqtwbisq
-  url: http://rabbit.dereferenced.org/~nenolod/distfiles/pkgconf-0.9.6.tar.bz2
+  url: https://github.com/pkgconf/pkgconf/archive/pkgconf-0.9.6.tar.gz
```
certik: You need to also update the `key`. Use `hit fetch https://github.com/pkgconf/pkgconf/archive/pkgconf-0.9.6.tar.gz` to get it.

johannesring:
The old link works fine for me.

dagss:
I think we should prefer original project links to the extent possible though, which I guess means the GitHub link in this case?

johannesring:
I'm not sure what is the original project link, but there is no `configure` script in the one from GitHub, so we will have to run `autogen.sh` to generate one.

dagss:
Ah.

johannesring:
[This](https://github.com/pkgconf/pkgconf/#release-tarballs) shows that we are already using the correct link.

agondolo:
The official tarball was down all of the workday yesterday. I had to download it manually and shove it into the .hashstack folder for the build to continue. And when I was looking around online, it seems to go down often.

dagss:
FWIW, you can download any (identical, same-hash) tarball from another 
location (including local disk) using `hit fetch` and then the listed 
location won't be tried during the build.

Issue #802, opened 2015-06-19
Boost fails to build with GCC 5


johannesring:
We got a [report](http://fenicsproject.org/pipermail/fenics/2015-June/002758.html) yesterday on the FEniCS mailing list about a problem with Boost ([build.log](https://gist.github.com/johannesring/e2fcd83a39a78d0759d5)). It turns out that the current Boost version in hashstack, 1.55.0, does not work with GCC 5 (also reported [here](https://bugs.freebsd.org/bugzilla/show_bug.cgi?id=197594)). Updating Boost to 1.56.0 or later will solve this problem as this is fixed [upstream](https://svn.boost.org/trac/boost/ticket/10125), however, for now I will create a pull request with the [patch](https://svn.boost.org/trac/boost/raw-attachment/ticket/10125/call_once_variadic.patch) that fixes the problem.

johannesring:
Fixed by #803.

PR #747, opened 2015-05-01
[WIP] Add gcc 5.1.0

certik:

PR #803, opened 2015-06-19
boost: add patch to fix problems with gcc5

johannesring:
This fixes issue #802.

### DIFF
--- a/pkgs/boost/boost.yaml
+++ b/pkgs/boost/boost.yaml
@@ -17,6 +17,13 @@ build_stages:
   bash: |
     ./bootstrap.sh --with-toolset={{toolset}} --prefix=${ARTIFACT}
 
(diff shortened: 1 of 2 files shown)


certik:
That's fine. But we can also just update Boost, isn't it easier?

johannesring:
Updating Boost is the better solution, but this requires more testing and we also have some other patches that needs attention if we update, so for now it was easier for me to just add the patch.

certik:
@johannesring  +1 to this PR then.

Not related to this PR, except using gcc 5, but I was wondering if you knew how to fix #747. It'd be nice to have gcc 5 in Hashstack itself.

johannesring:
Thanks @certik.

I'm not sure about #747, but I can take a look after the FEniCS workshop next week.

PR #804, opened 2015-06-25
cbcblock: remove trilinos as a required dependency

johannesring:


johannesring:
This does not affect other packages, so I'm merging it.

PR #805, opened 2015-06-27
boost: auto detect address-model

johannesring:
This PR adds auto detection of the address-model (32/64 bits) for Boost. The old parameter `address_model` will override the auto detection (maybe needed for cross compilation).


certik:
I've never heard of `getconf` before. But `getconf LONG_BIT` works on RHEL6 as well as on OS X 10.10, so I think this PR should work (I haven't tested).

+1

johannesring:
I  hadn't heard about it either. I was thinking about using `uname -m` but that does not work for instance inside a chroot, while `getopt LONG_BIT` does.

PR #808, opened 2015-07-07
Add the PyDy package

certik:
It works great in Jupyter.

@moorepants here is the build script for the PyDy package.


pkgs/pydy.yaml
```diff
@@ -0,0 +1,9 @@
+extends: [setuptools_package]
+
+dependencies:
+  build: [python]
+  run: [python, numpy, scipy, sympy, ipython, cython, matplotlib]
```
moorepants: Does "ipython" include the notebook?

moorepants: theano is also an optional dep like cython

certik: Yes, `ipython` includes the notebook. I still need to add theano to Hashstack.

PR #809, opened 2015-07-09
added joblib for simple parallel python

jenom:


pkgs/joblib.yaml
```diff
@@ -0,0 +1,7 @@
+# Joblib version 0.9.0b2 for simple prallel python loops.
+
+extends: [setuptools_package]
+
+sources:
+  - key: tar.gz:ggsl323ify27x6na4zsbbo76ealp7vis
+    url: https://pypi.python.org/packages/source/j/joblib/joblib-0.9.0b2.tar.gz#md5=530ead86cc3d07aa42d0db65adbc4dd8
```
johannesring: The `#md5=530ead86cc3d07aa42d0db65adbc4dd8` part of the `url` is not needed, but otherwise it looks good.

pkgs/joblib.yaml
```diff
@@ -0,0 +1,7 @@
+# Joblib version 0.9.0b2 for simple prallel python loops.
```
johannesring: Small typo here (missing a `a` in `prallel`).

pkgs/joblib.yaml
jenom: I will change it for the next release. BTW. is it preferred to use the latest git version or some fixed version like I did it here?

pkgs/joblib.yaml
jenom: Thanks. Is there anything else I have to do here like clicking on "close pull request" or is the rest automatic?

johannesring:
If you can fix the two minor issues I pointed out, then I can merge it. Nothing else you need to do. 

We usually prefer to have a release as default source unless there is a good reason to use the latest git version.

jenom:
changed minor and ready for merge. thanks for your patience. its my first time using github for bigger projects.

johannesring:
Great, thanks!

Issue #810, opened 2015-07-16
scotch is missing zlib in build dependencies


johannesring:
2015/07/15 14:32:52 - INFO: [package:run_job] running [u'/bin/bash', '_hashdist/build.sh']
    2015/07/15 14:32:52 - INFO: [package:run_job] environment:
    2015/07/15 14:32:52 - INFO: [package:run_job]   {'ARTIFACT': u'/root/.hashdist/bld/scotch/qaholoekwukv',
    2015/07/15 14:32:52 - INFO: [package:run_job]    'BASH': u'/bin/bash',
    2015/07/15 14:32:52 - INFO: [package:run_job]    u'BISON_DIR': u'/root/.hashdist/bld/bison/jri2642x26uh',
    2015/07/15 14:32:52 - INFO: [package:run_job]    u'BISON_ID': u'bison/jri2642x26uh23m4dmuubodq6wbimuib',
    2015/07/15 14:32:52 - INFO: [package:run_job]    'BUILD': u'/root/.hashdist/tmp/scotch-qaholoekwukv',
    2015/07/15 14:32:52 - INFO: [package:run_job]    u'FLEX_DIR': u'/root/.hashdist/bld/flex/eeduf457bgf4',
    2015/07/15 14:32:52 - INFO: [package:run_job]    u'FLEX_ID': u'flex/eeduf457bgf4ywsckky6ak5luhqckn2g',
    2015/07/15 14:32:52 - INFO: [package:run_job]    'HASHDIST_CPU_COUNT': '4',
    2015/07/15 14:32:52 - INFO: [package:run_job]    'HDIST_CONFIG': '{"gc_roots":"/root/.hashdist/gcroots","build_stores":[{"dir":"/root/.hashdist/bld"}],"source_caches":[{"dir":"/root/.hashdist/src"}],"cache":"/root/.hashdist/cache","build_temp":"/root/.hashdist/tmp"}',
    2015/07/15 14:32:52 - INFO: [package:run_job]    'HDIST_IMPORT': u'bison/jri2642x26uh23m4dmuubodq6wbimuib flex/eeduf457bgf4ywsckky6ak5luhqckn2g mpi/gmkx3ftk3tz2q6qreifxplulowzcim5h patchelf/k3rloj265ogtl4dmmmbmyt34dnffryka',
[...]

PR #811, opened 2015-07-16
scotch: add missing dependency on zlib

johannesring:
Fixes issue #810.

### DIFF
--- a/pkgs/scotch/Make.inc.linux
+++ b/pkgs/scotch/Make.inc.linux
@@ -9,9 +9,9 @@ CAT		= cat
 CCS		= $(MPICC)
 CCP		= $(MPICC)
 CCD		= $(MPICC)
-CFLAGS		= -fPIC -O3 -DCOMMON_FILE_COMPRESS_GZ -DCOMMON_PTHREAD -DCOMMON_RANDOM_FIXED_SEED -DSCOTCH_RENAME -Drestrict=__restrict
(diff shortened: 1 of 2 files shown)


pkgs/scotch/Make.inc.linux
```diff
@@ -9,9 +9,9 @@ CAT		= cat
 CCS		= $(MPICC)
 CCP		= $(MPICC)
 CCD		= $(MPICC)
-CFLAGS		= -fPIC -O3 -DCOMMON_FILE_COMPRESS_GZ -DCOMMON_PTHREAD -DCOMMON_RANDOM_FIXED_SEED -DSCOTCH_RENAME -Drestrict=__restrict
+CFLAGS		= -I$(ZLIB_DIR)/include -fPIC -O3 -DCOMMON_FILE_COMPRESS_GZ -DCOMMON_PTHREAD -DCOMMON_RANDOM_FIXED_SEED -DSCOTCH_RENAME -Drestrict=__restrict
 CLIBFLAGS	=
-LDFLAGS		= -lz -lm -lrt
+LDFLAGS		= -L$(ZLIB_DIR)/lib -lz -lm -lrt
```
certik: I think you also need to set the rpath here.

certik:
Otherwise +1.

pkgs/scotch/Make.inc.linux
johannesring: Thanks @certik, but does it matter when we build a static library?

certik: Ah, it doesn't matter at all for a static library, I didn't realize you were building a static library.

+1 to merge then.

johannesring:
Thanks!

PR #817, opened 2015-07-23
add proteus.tamucluster.yaml to stack/examples 

srobertp:
which gives a stack profile for users who wish to build on the numerical analysis Mathematics clusters located at texas a&m university.

### DIFF
--- a/examples/proteus.tamucluster.yaml
+++ b/examples/proteus.tamucluster.yaml
@@ -1,0 +1,72 @@
+# This profile file controls your <#> (HashDist) build environment.
+
+# This profile is written specific to the user srobertp.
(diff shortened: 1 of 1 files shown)


certik:
+1

cekees:
+1 

Just for future reference, it might be worth switching to the hashstack version of cmake eventually instead of relying on the /w/srobertp version of cmake via host-cmake.

srobertp:
@cekees good idea. That would make this .yaml file general for those who want to use it on the tamu cluster.

cekees:
This a proteus HPC  profile, so the test failures are unrelated.

PR #818, opened 2015-07-24
[WIP] Python on Cygwin64.

memmett:
This is close to working: some extension modules aren't being detected/compiled properly.  I think this may have something to do with how broken Windows is with respect to locating dynamic libraries (ie, no RPATH, just PATH).

Will do some more investigating now that I know about `cygcheck`.

### DIFF
--- a/pkgs/python/2.5.2-ctypes-util-find_library.patch
+++ b/pkgs/python/2.5.2-ctypes-util-find_library.patch
@@ -1,0 +1,34 @@
+--- Lib/ctypes/util.py	2007-09-14 15:05:26.000000000 -0500
++++ Lib/ctypes/util.py	2008-11-25 17:54:47.319296200 -0600
+@@ -41,6 +41,20 @@
+                 continue
+         return None
+ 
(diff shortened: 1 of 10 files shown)


certik:
+1. Did you figure out all these patches by yourself? Impressive work.

jcftang:
+1

memmett:
@certik Nope, I pulled in the patches from the cygwinports/python project...

cekees:
I'll try to test this out on a windows image at some point soon, but @memmett if you've got a stack that is working for you, then I'd vote for merging in all the cygwin support as soon as you're happy with it. Unless others are using the cygwin support and want to be more conservative. We don't have anybody currently using the proteus cygwin stack.

certik:
+1 to merge from me

memmett:
@cekees Alrighty, will do (sometime in the next few days).

memmett:
I built this again today and I'm happy with it.  The users Cygwin install should include: libcrypt-devel, libexpat-devel, libffi-devel.  All other dependencies are built by hashdist.

PR #819, opened 2015-07-24
OpenMM: Bump version and add patch for Cygwin.

memmett:
Compiles on Cygwin64 now.  Unfortunately users have to set their PATH appropriately so that Windows can find the `cygOpenMM*` DLLs (`cygcheck` is handy for this).

Maybe there is some `launcher` magic that we can do on Cygwin to populate the PATH appropriately?


certik:
+1, this looks good.

PR #824, opened 2015-08-22
Update netcdf4.yaml

odellus:
old url was invalid. project tarballs moved to github.


johannesring:
The netcdf4 test failed because the hash is not the same as the hash from the old url.

johannesring:
Also, the old url is not invalid.

certik:
closing in favor of #825

PR #825, opened 2015-08-22
Update netcdf4.yaml

odellus:
needed to change key as well as the URL for netcdf

### DIFF
--- a/pkgs/netcdf4/netcdf4.yaml
+++ b/pkgs/netcdf4/netcdf4.yaml
@@ -3,11 +3,7 @@ dependencies:
   build: [mpi, hdf5, curl]
 
 sources:
-<<<<<<< HEAD
-- key: tar.gz:xxpd3cyojdxnffeovvs7qlc47n2zamj3
-=======
 - key: tar.gz:6lxhr2zray34ab7qaht4ddrno46sh42f
->>>>>>> master
   url: https://github.com/Unidata/netcdf-c/archive/v4.3.3.1.tar.gz
 
 defaults:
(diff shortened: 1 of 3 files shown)


odellus:
Merged changes made to into one single PR.

jcftang:
looks good to me, it built for me, but I don't use netcdf4 myself to verify that it's functioning as it was before.

johannesring:
The old netcd4 url still works fine for me. Also, the old tarball is 4.8M, while the new is 17M. It is probably not important, but do you know why?

odellus:
No idea. It wasn't working yesterday. I noticed that none of the other
commits in the Travis builds were failing so the issue I was having seemed
to be local, but the old URL *was* failing for me in the script or when
checking the url with the browser.

I think some of the urls I added for netcdf-* might be to newer versions.
Feel free to close this PR w/o merge since it doesn't seem to be affecting
anyone else other than myself. :smile:

On Mon, Aug 24, 2015 at 12:41 AM, Johannes Ring <notifications@github.com>
wrote:

> The old netcd4 url still works fine for me. Also, the old tarball is 4.8M,
> while the new is 17M. It is probably not important, but do you know why?
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/hashdist/hashstack/pull/825#issuecomment-134074244>.
>



-- 
Sincerely,

Thomas O. Wood

certik:
The difference in size is due to the fact that the github version is just a tagged git repository, while the released tarball is presumably created by hand or by some script from it, probably omitting some files.

certik:
@odellus other PRs are merged, can you just create a clean PR for the netcdf?

jcftang:
This just needs some rebasing/removal of the merge conflict lines if I understand right.

johannesring:
It looks like everything in this PR has been merged, so I think this can be closed.

PR #826, opened 2015-08-23
Update netcdf4f.yaml

odellus:
new URL and key. Old URL was invalid.


jcftang:
Would it be more reasonable combine #825, #826 and #827 into on pulll request to make it easier to review. If its a project change with URL's it might be worth flattening the 3 patches. Other than that it looks good to me.

odellus:
Will do!

PR #830, opened 2015-08-25
Update netcdf4.yaml

odellus:
Clean PR containing only changes to netcdf4.yaml.

### DIFF
--- a/pkgs/netcdf4/netcdf4.yaml
+++ b/pkgs/netcdf4/netcdf4.yaml
@@ -3,8 +3,8 @@ dependencies:
   build: [mpi, hdf5, curl]
 
 sources:
-- key: tar.gz:xxpd3cyojdxnffeovvs7qlc47n2zamj3
-  url: ftp://ftp.unidata.ucar.edu/pub/netcdf/netcdf-4.3.3.1.tar.gz
+- key: tar.gz:6lxhr2zray34ab7qaht4ddrno46sh42f
+  url: https://github.com/Unidata/netcdf-c/archive/v4.3.3.1.tar.gz
 
 defaults:
   increase_max_dims: false


johannesring:
Does the old url still not work for you? It works fine for me:

    johannr@debian-t420s:src$ hit fetch ftp://ftp.unidata.ucar.edu/pub/netcdf/netcdf-4.3.3.1.tar.gz
    Downloading ftp://ftp.unidata.ucar.edu/pub/netcdf/netcdf-4.3.3.1.tar.gz...
    [=========================] 100.0% (4.8MB of 4.8MB) 2.236MB/s ETA 0s  

    sources:
    - key: tar.gz:xxpd3cyojdxnffeovvs7qlc47n2zamj3
      url: ftp://ftp.unidata.ucar.edu/pub/netcdf/netcdf-4.3.3.1.tar.gz

What is the error message you get? Can you also try using `wget`?

I am ok with merging this if the old url does not work for you.

odellus:
@johannesring I received a `500 error (could not change directory)` or something to that effect the other day both when using `wget` through the proteus build script as well as when entering the ftp location into Firefox. @certik was having a similar issue and suggested a clean PR w/ just this single change.

The old URL seems to be working fine for me too right now, so disregarding this change wouldn't be a bad idea. It could be opened and merged at a later time easily if the problem occurs again with any frequency. :smile:

johannesring:
If it was just a temporary problem, and it does not happen too often, then I agree that it is better to keep the old URL for now.

cekees:
A student and  I both had a problem with this URL about a month ago, but I was visiting another  lab and assumed I was just having trouble  with the guest network. I think it might be good to go ahead and merge this. Thanks for  making the change @odellus .

johannesring:
OK, then I'm +1.

jcftang:
+1

ahmadia:
+1

PR #831, opened 2015-09-03
Update fabric to 1.10.2 stable

jcftang:
This brings the fabric package up to date

### DIFF
--- a/pkgs/fabric.yaml
+++ b/pkgs/fabric.yaml
@@ -4,5 +4,23 @@ dependencies:
   run: [paramiko]
 
 sources:
-  - url: https://pypi.python.org/packages/source/F/Fabric/Fabric-1.8.0.tar.gz
-    key: tar.gz:ab3mo2v654775els4euq43sipyr2oiu3
+- key: tar.gz:mzvd3kjconjs2osj3unmkpr5oiof5qxs
+  url: https://pypi.python.org/packages/source/F/Fabric/Fabric-1.10.2.tar.gz
+
+
+build_stages:
+
+# Due to this error:
+#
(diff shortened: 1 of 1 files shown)


pkgs/fabric.yaml
```diff
@@ -4,5 +4,16 @@ dependencies:
...
-    key: tar.gz:ab3mo2v654775els4euq43sipyr2oiu3
+- key: tar.gz:mzvd3kjconjs2osj3unmkpr5oiof5qxs
+  url: https://pypi.python.org/packages/source/F/Fabric/Fabric-1.10.2.tar.gz
+
+build_stages:
+
+- name: install
+  after: setup_dirs
+  handler: bash
+  bash: |
```
ahmadia: should this be a setuptools or a distutils install?

jcftang: It was failing to create the profile due to a write permission error, the build completes if I build/install manually in this profile.

ahmadia: huh, weird.  maybe put a comment in the file explaining what's going on for the next person who looks at this.

jcftang: Done, I've amended the commit

jcftang:
thanks!

pkgs/fabric.yaml
certik: What *exactly* is the problem here? This seems to be fixing the symptoms, without understanding the root of the cause.

jcftang: This was the error at profile generation...

[profile|ERROR] hit command failed: [Errno 13] [Errno 13] Permission denied: '/Users/jtang/.hashdist/bld/profile/t63wczbqichu/lib/python2.7/site-packages/tests/etree13' in silent_makedirs(u'/Users/jtang/.hashdist/bld/profile/t63wczbqichu/lib/python2.7/site-packages/tests/etree13',)

certik: So is the problem that the package is installing a `tests` directory directly into the Python's package `site-packages`? And why is this avoided by your patch? What is different?

jcftang: To be honest, I didn't spend much time investigating, it was what worked for me at the time. I think this has arisen a few times for me in the past, I previously just swapped out setuptools for distutils or vice versa and the issue goes away.

PR #836, opened 2015-09-16
Add sgi icex support

cekees:
Cherry picked  support for SGI ICEX from stable/spirit branch. This shouldn't break  anything for anyone else. Note: stable/spirit still has changes due to  SSL issues on that machine that required using alternative URL's for  some of the sources. Once  those issues go away, commits in this PR  should be all that is needed.

### DIFF
--- a/pkgs/numpy/numpy.yaml
+++ b/pkgs/numpy/numpy.yaml
@@ -30,6 +30,29 @@ build_stages:
       export LAPACK=None
       export BLAS=None
 
+  - when: machine == 'SGIICEX'
+    name: create-site.cfg
+    after: prologue
+    before: install
+    handler: bash
+    bash: |
+      cat > site.cfg << EOF
+      [mkl]
+      library_dirs = /app/intel/compilers/14/mkl/lib/intel64
+      mkl_libs = mkl_rt,pthread,m
+      EOF
+
+  - when: machine == 'SGIICEX'
+    name: set-LDFLAGS
+    after: prologue
+    before: install
+    handler: bash
+    bash: |
+      export LDFLAGS="-shared -Wl,-rpath=${PYTHON_DIR}/lib $(${PYTHON_DIR}/bin/python-config --ldflags)"
+      export LAPACK=None
+      export BLAS=None
+      export ATLAS=None
+
   - when: machine == 'CrayXE6'
     name: create-site.cfg
     after: prologue
--- a/pkgs/daetk/daetk.yaml
+++ b/pkgs/daetk/daetk.yaml
@@ -26,6 +26,24 @@ build_stages:
     export MPI=$MPI_DIR
     touch dep.txt
 
+- when: machine == 'SGIICEX'
+  name: configure
+  mode: replace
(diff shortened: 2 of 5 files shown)


pkgs/numpy/numpy.yaml
```diff
@@ -30,6 +30,29 @@ build_stages:
...
 
+  - when: machine == 'SGIICEX'
+    name: create-site.cfg
+    after: prologue
+    before: install
+    handler: bash
+    bash: |
+      cat > site.cfg << EOF
+      [mkl]
+      library_dirs = /app/intel/compilers/14/mkl/lib/intel64
```
certik: I think that this should not be in here. Rather, this should be specified in the user profile for this particular machine.

I am ok with merging this as a temporary measure, but this is not the way to go in the long term.

cekees: Here I'm just following what was  done  for other platforms (linux, cygwin, CrayXE6).  How  do  you suggest that we refactor the platform specific  info?

certik: This is about linking Lapack and Blas. We've discussed this at length in some other issue, I think the way forward is to export Hashstack variables about where to find the given blas, and then in the user profile enable it. It's sort of like that already, but it's not finished. Then from here, i.e. form `numpy.yaml`, you just use something like `library_dirs=${LAPACK_DIRS}` or something like that.

certik:
As I said, this continues the current approach in `numpy.yaml`, so we can merge it, but this is not the way in the long term. This is obvious, since we want the packages in the `pkgs` directory to work on all platforms, and the path `/app/intel/compilers/14/mkl/lib/intel64` only works on one particular computer, so this path does not belong into `pkgs/*`, but rather into a user profile for this particular computer.

cekees:
Thanks. I agree that we  need to continue to work out the lapack/blas parameter issue referenced  elsewhere.  I disagree that the information you flagged should go into a user profile. HPC machines are platforms unto themselves, and it's important for the community that  we maintain this platform-specific  configuration information as part of hashstack  in a way that is  easy for others to reuse.  For example, somebody building custom stacks on a Cray XE6 or SGI ICE X shouldn't have to go searching through  proteus profiles (or even worse, some fork of hashstack) just to figure out how to build a stack with python + scipy. It ought to  just work  if somebody has done it before on that particular HPC machine and be highly likely to work it it's a related machine of the same platform (e.g. the handful of XE6's at DoD HPC centers the handful of Q's at DoE centers, etc.).

certik:
I think we need some kind of machine files for Cray XE6 or SGI ICE X, and so far I envisioned those to be the user profiles. If you think that Cray XE6 is a platform that is always the same, i.e. perhaps like linux or OS X, then we can have these paths hardwired in the packages themselves. Even then though, I would like the Lapack paths to be in the Lapack packages, not in the numpy package.

cekees:
I see. I'll think about it. I  definitely agree that numpy.yaml should be getting this from the lapack  and blas packages specified by the profile. Better support  for modules (as @dagss has outlined) will also  go a long way to eliminating some of these machine/platform-specific configurations.

PR #841, opened 2015-09-16
added  pythreejs pkg

cekees:
requires #839

### DIFF
--- a/pkgs/pythreejs.yaml
+++ b/pkgs/pythreejs.yaml
@@ -1,9 +1,9 @@
 extends: [setuptools_package]
 
 dependencies:
-  build: [ipython,pip,jupyter-pip]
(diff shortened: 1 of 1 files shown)


jcftang:
+1

certik:
I fixed the style. +1 now.

cekees:
@certik it looks like the checks fails just because of random slow tests taking over 50min. Have you just been merging PR's in this case? I don't see any way that this pkg could cause a test failure.

johannesring:
The dolfin test is failing for some reason and it is not because of a timeout. I tried to force a rebuild but it failed at the same place and I don't see why because we only see the last 60 lines of `log.txt`. I don't see why adding this package should cause any problems, so I am fine with merging this.

certik:
Yes, I was merging it if it goes over 50min.

PR #842, opened 2015-09-16
updated versions and url's to github, enabled core and  image

cekees:
@johannesring I'm putting this up in case it is of use.  I don't use cgal much, and dolfin is the only other stack I know that  might use  it.  It's basically just an update to a more recent version of cgal and a reflection of the switch from google code to  github for the python bindings hosting.

### DIFF
--- a/pkgs/cgal.yaml
+++ b/pkgs/cgal.yaml
@@ -3,8 +3,8 @@ dependencies:
   build: [zlib, boost, gmp, mpfr]
 
 sources:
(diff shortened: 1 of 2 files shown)


johannesring:
@cekees - I am fine with this change. DOLFIN does not use CGAL anymore after the mesh generation part of DOLFIN was moved to mshr, which again comes bundled with its own copy of CGAL.

certik:
+1 from me.

PR #843, opened 2015-09-17
python: update to 2.7.10

jcftang:
This is a periodic update.

2.7.10 built cleanly for me and the dependencies that I use.


johannesring:
+1

PR #847, opened 2015-09-26
updated petsc4py and  daetk to work with petsc v3.6.1

cekees:
@johannesring I tested the petsc 3.6.1 version you added, and it is working fine for me. This PR has some of the changes petsc4py and daetk that  I ended up making to run proteus with the new version.

### DIFF
--- a/pkgs/daetk/arch.cygwin
+++ b/pkgs/daetk/arch.cygwin
@@ -1,4 +1,4 @@
-include ${PETSC}/conf/variables
+include ${PETSC}/lib/petsc/conf/variables
 PREFIX=${ARTIFACT}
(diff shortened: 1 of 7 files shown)


certik:
+1 from me.

johannesring:
Is the patch for petsc4py something that should be applied upstream?

Otherwise +1 from me.

cekees:
Good idea.  I'll discuss  it  upstream.

ahmadia:
@cekees : see https://bitbucket.org/petsc/petsc4py/issues/2/consider-adding-conf-to-petsc4py-install

cekees:
OK, thanks. I see his  point. I can take a look at getting the info I need from somewhere else.

PR #849, opened 2015-09-27
Add jupyter

cekees:
This still needs some cleanup and dependency checking, but I thought I'd go ahead and put it up in case others have suggestions or  concerns.


certik:
That looks good.

minrk:
Looks sensible, but most of the IPython/Jupyter packages are missing dependencies.

ahmadia:
Have all of these switched to setuptools?  How annoying.

cekees:
That's a fair question. Actually, it looks like the skeleton-pypi command has setuptools hardwired  into the template and that the auto-discovery of dependencies has been left as a todo item.  It may be that most of the packages in this PR could be distutils packages. I don't have time this week, but jupyter might be a good test case for a refactored/extended skeleton-pypi subcommand.

minrk:
@ahmadia none of the IPython or Jupyter packages need setuptools to install. I'm not sure how relevant that is to hashdist, or what hashdist does differently for setuptools vs distutils packages. I would guess that ~all Python packages ought to be installed with `pip install .`, regardless of whether the default behavior is to import setuptools or not.

ahmadia:
We use `python setup.py install` with just `distutils` in the environment for `distutils` packages, and something slightly hairier for `setuptools`.  Generally it's a better idea to avoid adding `setuptools` as an explicit dependency.  Using `pip install .` may be feasible, I've never looked into it.

cekees:
It appears that jupyter needs appnope  and gnureadline on os x as well. 
/cc @guoyinwang

certik:
I fixed it up to work for me and pushed the patches into this PR. I think the dependencies should be more or less correct.

@cekees any objections against merging this? We can always improve on it further once it is in.

certik:
If this works, then let's merge. I am updating to the latest release in #932, but that has its own issues, which are better handled by a separate PR.

cekees:
Sounds good.  I'll wait until the checks complete.

certik:
Checks completed and passed. Merging.

PR #851, opened 2015-10-02
Use http for openssl download

vbraun:
Also update to latest openssl release

Fixes https://github.com/hashdist/hashdist/issues/352


cekees:
+1. Thanks for identifying hashdist/hashdist#352 and fixing it. and also for going to the trouble to add the docs patch.

certik:
+1 as well, I just merged it.

In the long run we probably want to allow some more general way to get the sources, since we don't care how the tarball is obtained and if it the connection/download can be trusted, since we check the checksum, which should be cryptographically secure.

PR #855, opened 2015-10-05
Update to Ruby-2.2.3

vbraun:
This fixes a bug in ruby where it couldn't connect to its own server
due to a SSL/TLS incompatibility:

$ hit develop
$ ./default/bin/gem --install-dir ./default install bundler
ERROR:  Could not find a valid gem 'bundler' (>= 0), here is why:
          Unable to download data from https://rubygems.org/ - SSL_connect returned=1 errno=0 state=error: certificate verify failed (https://api.rubygems.org/latest_specs.4.8.gz)

See also: https://github.com/rubygems/rubygems/issues/665


certik:
That works, thanks!

jcftang:
+1

Issue #856, opened 2015-10-05
gccxml fails to build with gcc 5.1


johannesring:
This is the error I get when compiling with gcc 5.1:

``` console
[gccxml] [ 72%] Building C object GCC/gcc/CMakeFiles/backend.dir/toplev.c.o
[gccxml] [ 72%] Building C object GCC/gcc/CMakeFiles/backend.dir/tracer.c.o
[gccxml] [ 72%] Building C object GCC/gcc/CMakeFiles/backend.dir/tree.c.o
[gccxml] [ 73%] Building C object GCC/gcc/CMakeFiles/backend.dir/tree-dump.c.o
[gccxml] /cluster/home/johannr/.hashdist/tmp/gccxml-twxkdnzswsnv-1/GCC/gcc/toplev.c:546:1: error: redefinition of 'floor_log2'
[gccxml]  floor_log2 (unsigned HOST_WIDE_INT x)
[gccxml]  ^
[gccxml] In file included from /cluster/home/johannr/.hashdist/tmp/gccxml-twxkdnzswsnv-1/GCC/gcc/toplev.c:59:0:
[gccxml] /cluster/home/johannr/.hashdist/tmp/gccxml-twxkdnzswsnv-1/GCC/gcc/toplev.h:174:1: note: previous definition of 'floor_log2' was here
[gccxml]  floor_log2 (unsigned HOST_WIDE_INT x)
[gccxml]  ^
[gccxml] /cluster/home/johannr/.hashdist/tmp/gccxml-twxkdnzswsnv-1/GCC/gcc/toplev.c:581:1: error: redefinition of 'exact_log2'
[gccxml]  exact_log2 (unsigned HOST_WIDE_INT x)
[gccxml]  ^
[gccxml] In file included from /cluster/home/johannr/.hashdist/tmp/gccxml-twxkdnzswsnv-1/GCC/gcc/toplev.c:59:0:
[gccxml] /cluster/home/johannr/.hashdist/tmp/gccxml-twxkdnzswsnv-1/GCC/gcc/toplev.h:180:1: note: previous definition of 'exact_log2' was here
[gccxml]  exact_log2 (unsigned HOST_WIDE_INT x)
[gccxml]  ^
[gccxml] gmake[2]: *** [GCC/gcc/CMakeFiles/backend.dir/toplev.c.o] Error 1
[gccxml] gmake[2]: *** Waiting for unfinished jobs....
[...]

PR #857, opened 2015-10-05
gccxml: add patch to fix build problems with gcc 5.1

johannesring:
Fixes issue #856.


certik:
That looks good, thanks!

johannesring:
Thanks!

Issue #593, opened 2014-12-15
FEniCS linking broken on Darwin


alogg:
When building FEniCS with HashStack on Mac OS X Yosemite 10.10.1, I get the following error when trying to import a FEniCS Python module after a seemingly successful build with HashStack/HashDist:

    $ python -c "import dolfin"
    Traceback (most recent call last):
    File "<string>", line 1, in <module>
    File "/Users/logg/.hashdist/bld/profile/3cy7e2il6iie/lib/python2.7/site-packages/dolfin/__init__.py", line 16, in <module>
    from . import cpp
    File "/Users/logg/.hashdist/bld/profile/3cy7e2il6iie/lib/python2.7/site-packages/dolfin/cpp/__init__.py", line 42, in <module>
    exec("from . import %s" % module_name)
    File "<string>", line 1, in <module>
    File "/Users/logg/.hashdist/bld/profile/3cy7e2il6iie/lib/python2.7/site-packages/dolfin/cpp/common.py", line 32, in <module>
    _common = swig_import_helper()
    File "/Users/logg/.hashdist/bld/profile/3cy7e2il6iie/lib/python2.7/site-packages/dolfin/cpp/common.py", line 28, in swig_import_helper
    _mod = imp.load_module('_common', fp, pathname, description)
    ImportError: dlopen(/Users/logg/.hashdist/bld/profile/3cy7e2il6iie/lib/python2.7/site-packages/dolfin/cpp/_common.so, 2): Library not loaded: libdolfin.1.4.dylib
    Referenced from: /Users/logg/.hashdist/bld/profile/3cy7e2il6iie/lib/python2.7/site-packages/dolfin/cpp/_common.so
    Reason: image not found

johannesring:
Fixed in #861.

PR #861, opened 2015-10-19
rpath fixes for os x

johannesring:
In the latest version of OS X (10.11), setting `DYLD_LIBRARY_PATH` and `DYLD_FALLBACK_LIBRARY_PATH` does not work as in earlier versions of OS X. Using rpath is therefore important. This PR fixes this problem for the FEniCS stack (#593).

### DIFF
--- a/pkgs/boost/boost.yaml
+++ b/pkgs/boost/boost.yaml
@@ -117,7 +117,7 @@ build_stages:
   handler: bash
   bash: |
     for boostlib in ${ARTIFACT}/lib/libboost_*.dylib; do
-        install_name_tool -add_rpath ${ARTIFACT}/lib/ ${boostlib}
+        install_name_tool -add_rpath @loader_path ${boostlib}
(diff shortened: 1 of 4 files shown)


ahmadia:
+1, it almost looks like this could be made into an inheritable function hook eventually.

johannesring:
Yes, that would be good. 

BTW, do you know how to make the rpath relocatable? Can we use `@loader_path` or something like that? For Boost I just added the full path to the `lib` directory.

ahmadia:
Yeah, something like @loader_path should work.

johannesring:
Do you think the following will work (for Boost)? 

    install_name_tool -add_rpath @loader_path ${boostlib}

Or do we need to add something after `@loader_path`? I will try it tomorrow.

johannesring:
Using `@loader_path` worked fine.

cekees:
+1

PR #864, opened 2015-11-02
[WIP] Added support for ARCHER (Cray XC30).

cekees:
Not quite ready to  go in

### DIFF
--- a/examples/proteus.archer.yaml
+++ b/examples/proteus.archer.yaml
@@ -1,0 +1,75 @@
+# This profile file controls your <#> (HashDist) build environment.
+
+# In the future, we'll provide better incorporation of
+# automatic environment detection.  For now, have a look
+# at the YAML files in the top-level directory and choose
+# the most *specific* file that matches your environment.
+
+extends:
+- file: archer.yaml
+
+# The packages list specifies all the packages that you
+# require installed.  <#> will ensure that all packages
+# and their dependencies are installed when you build this
+# profile.
+
+parameters:
+  machine: CrayXC30
+  HOST_MPICC: cc
+  HOST_MPICXX: CC
+  HOST_MPIF77: ftn
+  HOST_MPIF90: ftn
+  HOST_MPIEXEC: aprun
+  PATH: |
+    /usr/bin:/bin:/usr/sbin:/sbin
+  PROLOGUE: |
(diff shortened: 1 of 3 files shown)


examples/proteus.archer.yaml
```diff
@@ -0,0 +1,75 @@
...
+    use: host-mpi
+  mpi4py:
+  nose:
+  hdf5:
+  ipython:
+  matplotlib:
+  parmetis:
+    without_check: true
+  petsc:
+    build_with: |
```
cekees: we should update to the latest petsc version by adding `version: 3.6.1`  above `without_check: true`

examples/proteus.archer.yaml
```diff
@@ -0,0 +1,75 @@
...
+      parmetis
+    download: |
+      ptscotch, superlu, superlu_dist, spooles, hypre, blacs, scalapack, mumps
+    coptflags: -O2
+    link: shared
+    debug: false
+  petsc4py:
+    with_conf: true
```
cekees: petsc4py currently needs  `version: 3.6.0` to  work with petsc version 3.6.1. (The petsc4py 3.6.1 was not  out yet last time I checked).

pkgs/hdf5/hdf5.yaml
```diff
@@ -18,15 +18,16 @@ build_stages:
   bash: |
     export CC=${MPICC}
 
-- when: machine == 'CrayXE6'
+- when: machine == 'CrayXC30'
```
cekees: The hashdist's YAML domain specific  langauge understands basic logic, so if you make this `when: machine == 'CrayXE6' or machine == 'CrayXC30'`. Your changes will be backward compatible with the existing XE6 support.

cekees:
@mbareford and @tridelat I will try these out on our XC30 once we're able to build the whole stack again on Archer.

examples/proteus.archer.yaml
mbareford: Thanks, I've made the changes you recommended and daetk now compiles.

pkgs/hdf5/hdf5.yaml
```diff
@@ -18,15 +18,16 @@ build_stages:
...
 
-- when: machine == 'CrayXE6'
+- when: machine == 'CrayXE6' or machine == 'CrayXC30'
   name: make_exe_builds_static
   handler: bash
   after: configure
   before: make
   bash: |
-    for f in {test,testpar,tools/*,perform,hl/*,hl/*/*}/Makefile 
+    for f in {src,test,testpar,tools/*,perform,hl/*,hl/*/*}/Makefile
```
cekees: @mbareford this causes a runtime problem with the hdf5 shared libraries when I test  on our XC30. The issue is that the '-static' option added in the original patch was  only intended for  the directories in the original list so it  breaks the shared  libs at runtime when it is also applied  to src. I made a branch fix_hdf5_sed that applies your sed fix in a second  loop and verified  that it  works. Could you merge that into your  branch and test on Archer? 
/cc @tridelat @adimako

PR #868, opened 2015-11-03
fix typo on openblas yaml causing unbounded parallel make

### LINKED ISSUE #867: openblas parallel build typo and incorrect core count
@mbareford reported that the openblas build was asking for more processors than available on the login nodes of an HPC platform. I was looking at the opeblas pkg, and there appears to be a typo (HASDIST):
https://github.com/hashdist/hashstack/blob/master/pkgs/openblas.yaml#L34

I also would have thought that hashdist does serial builds of the `-j` option is not given.  Does it guess the number or cores?

cekees:
@mbareford and @tridelat, I think this should fix the archer build on login nodes. closes #867

### DIFF
--- a/pkgs/openblas.yaml
+++ b/pkgs/openblas.yaml
@@ -31,7 +31,7 @@ build_stages:
   after: prologue
   handler: bash
   bash: |
-    make USE_THREAD=${USE_THREAD} {{extra_flags}} -j${HASDIST_CPU_COUNT}
+    make USE_THREAD=${USE_THREAD} {{extra_flags}} -j${HASHDIST_CPU_COUNT}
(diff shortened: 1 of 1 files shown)


johannesring:
+1

certik:
+1, I completely missed this typo when I wrote the build script.

certik:
Or I guess @johannesring added it. Either way, several people have modified it and nobody noticed that it doesn't work. :)

johannesring:
Yes, I am the one to blame for the typo, but the thing is that the `-j` option work fine without a number when building OpenBLAS. It will only run two to five jobs simultaneously on my laptop with four cores. If I run `make -j` for DOLFIN, it will run "unlimited" simultaneous jobs and my laptop runs out of resources pretty fast.

Issue #866, opened 2015-11-03
golang1.4 is does not build on osx 10.11


jcftang:
I recently upgraded to the latest 10.11 release of osx and noticed that golang1.4 no longer builds correctly.

It maybe worth changing golang1.4 to use binaries. golang1.4 is just used to bootstrap golang1.5.

PR #874, opened 2015-11-07
Update golang and pycryptopp

jcftang:
This fixes https://github.com/hashdist/hashstack/issues/866 and also makes pycryptopp more portable.


jcftang:
This is safe to merge, the failure is not related to this.

PR #877, opened 2015-11-09
Update to nodejs v5

vbraun:
Straightforward update


vbraun:
This should be pretty uncontroversial, merging...

PR #879, opened 2015-11-10
Rethinkdb

vbraun:
This is the database that powers sagemathcoud. It depends on the protobuf update from #878


vbraun:
There are library linking issues, it picks up unwanted system-wide libraries for me

vbraun:
Now uses hashdist libraries except for the ones dragged in by curl https://github.com/hashdist/hashstack/issues/880

cekees:
+1

certik:
+1

Issue #886, opened 2015-11-18
Building zmq with Intel compiler fails


johannesring:
Relevant part of `build.log`:
``` console
[zmq] Making all in src
[zmq] make[1]: Entering directory `/cluster/home/johannr/.hashdist/tmp/zmq-dlk3fsbe5baa/src'
[zmq] make  all-am
[zmq] make[2]: Entering directory `/cluster/home/johannr/.hashdist/tmp/zmq-dlk3fsbe5baa/src'
[zmq]   CXX    libzmq_la-address.lo
[zmq]   CXX    libzmq_la-clock.lo
[zmq]   CXX    libzmq_la-ctx.lo
[zmq]   CXX    libzmq_la-decoder.lo
[zmq]   CXX    libzmq_la-devpoll.lo
[zmq]   CXX    libzmq_la-dist.lo
[zmq]   CXX    libzmq_la-encoder.lo
[zmq]   CXX    libzmq_la-epoll.lo
[zmq]   CXX    libzmq_la-err.lo
[zmq]   CXX    libzmq_la-fq.lo
[zmq]   CXX    libzmq_la-io_object.lo
[zmq]   CXX    libzmq_la-io_thread.lo
[zmq]   CXX    libzmq_la-ip.lo
[zmq]   CXX    libzmq_la-ipc_address.lo
[zmq]   CXX    libzmq_la-ipc_connecter.lo
[zmq] In file included from ipc_connecter.cpp(28):
[zmq] stream_engine.hpp(97): error #82: storage class is not first
[zmq]           const static size_t greeting_size = 12;
[zmq]                 ^
[zmq] 
[zmq] compilation aborted for ipc_connecter.cpp (code 2)
[zmq] make[2]: *** [libzmq_la-ipc_connecter.lo] Error 1
[zmq] make[1]: *** [all] Error 2
[zmq] make[2]: Leaving directory `/cluster/home/johannr/.hashdist/tmp/zmq-dlk3fsbe5baa/src'
[zmq] make[1]: Leaving directory `/cluster/home/johannr/.hashdist/tmp/zmq-dlk3fsbe5baa/src'
[zmq] make: *** [all-recursive] Error 1
[zmq|ERROR] Command '[u'/bin/bash', '_hashdist/build.sh']' returned non-zero exit status 2
[zmq|ERROR] command failed (code=2); raising
```
Upgrading to the latest version of zmq fixes this problem, but this requires other changes as well.

johannesring:
[...]

PR #887, opened 2015-11-18
Update zmq and pyzmq to fix problem with Intel compilers

johannesring:
Fixes issue #886.


johannesring:
I forgot to mention that the patch for pyzmq was made by taking [these](https://bitbucket.org/mpi4py/mpi4py/src/259f09/conf/mpidistutils.py?at=master&fileviewer=file-view-default#mpidistutils.py-40:51) lines from mpi4py.

ahmadia:
+1

PR #890, opened 2015-12-02
modelparameters: update to latest commit and switch to setuptools

johannesring:


johannesring:
Should be fine to merge.

PR #891, opened 2015-12-02
Add package cbcbeat

johannesring:


johannesring:
This is also ok to merge.

PR #894, opened 2015-12-04
Add TeX Live distribution

certik:
This only installs the binaries. To get a usable TeX distribution, one has to
also install the supporting files that are maintained separately.

### DIFF
--- a/pkgs/texlive.yaml
+++ b/pkgs/texlive.yaml
@@ -1,0 +1,24 @@
+extends: [base_package]
+
+sources:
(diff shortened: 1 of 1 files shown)


certik:
This builds in about 8 min with `-j16` on my machine, which is not too bad. As mentioned in the commit message, to get a working TeX distribution, we need to figure out how to get the supporting files working, which are distributed separately by upstream. That actually might be the hard part. But this is a good start.

jcftang:
+1

PR #895, opened 2015-12-06
Update to openssl 1.0.2e

vbraun:


vbraun:
Should be uncontroversial...

johannesring:
+1

PR #902, opened 2015-12-17
OS X fixes

certik:
Few fixes to get things working for El Capitan.

### DIFF
--- a/base/cmake_package.py
+++ b/base/cmake_package.py
@@ -18,6 +18,7 @@ def configure(ctx, stage_args):
           set_env_flags: true # default
           build_in_source: false # default
           debug: false # default
+          empty_osx_deployment_target: false # default
 
     If set_env_flags is set, CPPFLAGS and LDFLAGS will be set, as appropriate
     for the platform.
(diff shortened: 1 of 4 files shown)


certik:
@ahmadia, @johannesring would you mind reviewing this?

johannesring:
I have used a [similar approach](https://bitbucket.org/fenics-project/fenics-developer-tools/src/9beeb72723e60fe8215b7e27805562098680ec4f/install/patches/hashstack-cmake-osx-sysroot.patch) for the FEniCS installation script. That is, leaving `MACOSX_DEPLOYMENT_TARGET` untouched and setting `CMAKE_OSX_SYSROOT` to `/`. I think the latter is needed at least on OS X 10.10, because the latest Xcode only comes with

johannesring:
... because the latest Xcode only comes with 10.11 SDK. I don't remember exactly now unfortunately.

certik:
Thanks. It works for me, so I am going to merge this.

base/cmake_package.py
```diff
@@ -34,7 +35,7 @@ def configure(ctx, stage_args):
     else:
         conf_lines.append('-DCMAKE_BUILD_TYPE:STRING=Release')
 
-    if ctx.parameters['platform'] == 'Darwin':
+    if stage_args.get('empty_osx_deployment_target', False):
```
johannesring: @certik - I think this change also requires a change in `cmake_package.yaml` because setting `empty_osx_deployment_target: true` has currently no effect. I think what is missing is this:

``` diff
diff --git a/base/cmake_package.yaml b/base/cmake_package.yaml
index 3cd08de..7464694 100644
--- a/base/cmake_package.yaml
+++ b/base/cmake_package.yaml
@@ -15,6 +15,7 @@ build_stages:
   - name: configure
     after: setup_builddir
     debug: {{debug}}
+    empty_osx_deployment_target: {{empty_osx_deployment_target}}
 
   - name: make
     after: configure
```

ahmadia:
Sorry I missed this.  It looks good.

PR #905, opened 2016-01-05
mshr: make vtk optional

johannesring:


johannesring:
This is straightforward - merging.

PR #913, opened 2016-01-21
Add starcluster

johannesring:
This adds the StarCluster package and all the dependencies that were not already available in HashStack. It also updates paramiko to latest version 1.16.0 (required by StarCluster) and adds ecdsa as dependency to paramiko.


jcftang:
+1 this looks good

johannesring:
Thanks @jcftang.

PR #916, opened 2016-01-21
Update cmake to 3.4.2

certik:


johannesring:
I have tested this with my FEniCS profiles and it fails for mshr. Maybe we should have a `version` parameter for CMake like we have for PETSc?

jcftang:
@johannesring what's the error?

johannesring:
@jcftang - The error happens when mshr tries to build CGAL (mshr includes its own copy of CGAL). I don't remember exactly what the error was, but it is not important since the problem has been fixed in the development version of mshr by updating CGAL. I'm okay with merging this, but it will require an update of the mshr package. That should be fine since the development version of mshr should work with the stable version of the other FEniCS packages.

certik:
If it was fixed in some of those packages, then I think we should just merge this and then update those. I am merging this. @johannesring let me know if you need help with the `mshr` package.

PR #918, opened 2016-01-26
fiat: switch from distutils to setuptools

johannesring:


johannesring:
The failure is not related so merging.

Issue #919, opened 2016-01-27
flufl_lock fails on OS X


johannesring:
After 00c67e331878c0982436c7a5a5f7d2ec0b87fdf8 I now get the following error on OS X:

``` console
2016/01/27 15:24:11 - INFO: [package:run_job] cp: directory /Users/johannr/.hashdist/bld/flufl_lock/brgapi6mie2f/Python.framework/Versions/2.7/lib/python2.7/site-packages/flufl does not exist
2016/01/27 15:24:11 - ERROR: [package:run_job] Command '[u'/bin/bash', '_hashdist/build.sh']' returned non-zero exit status 1
2016/01/27 15:24:11 - ERROR: [package:run_job] command failed (code=1); raising
```

johannesring:
Fixed in #920.

PR #920, opened 2016-01-28
flufl_lock: copy __init__.py to correct location also on OS X

johannesring:
This will fix issue #919.


johannesring:
All tests pass. Merging.

PR #921, opened 2016-02-12
Add package pytest-cov

johannesring:
This also adds host packages for setuptools, pytest and pytest-cov.


johannesring:
This doesn't change anything, so I'm merging this.

PR #926, opened 2016-02-16
Fix openjpeg version to 1.5

robertsawko:
The symbolic link created by the fix_include step was hard-coded and
pointing to a different version. Also, version 2.1 was breaking grib
compilation.

### DIFF
--- a/pkgs/openjpeg.yaml
+++ b/pkgs/openjpeg.yaml
@@ -10,8 +10,8 @@ defaults:
   relocatable: false
 
 sources:
-- key: tar.gz:cizlxakp3cgy5uyuzfhqx7v3appikwky
-  url: http://sourceforge.net/projects/openjpeg.mirror/files/2.1.0/openjpeg-2.1.0.tar.gz
+- key: tar.gz:zgii7k2eimvc5usvbrcxzn7hg3ng27bs
+  url: http://sourceforge.net/projects/openjpeg.mirror/files/openjpeg-1.5.0.tar.gz
 
 build_stages:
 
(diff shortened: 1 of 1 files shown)


cekees:
This is on the stable/archlinux branch that I maintain, so I'd like to merge this. The test failures are due to a mismatch with the hashdist version. I'll look into whether I can fix that.

robertsawko:
Hi Chris,

This is really just a quick question rather than an issue. Is there a
way to invoke "calculateSolution" in parallel from within python? You
already pointed me to a relevant, I think, test file which is 

tests/ci/test_meshPartitionFromTetgenFiles.py

I am working with it now and trying to adjust my serial cases for
instance this one:

https://github.com/robertsawko/pde-and-uq/blob/master/riemann/mesh_and_discretisation_test.py

Could you give me a description of what needs to be done to make it work
in parallel?

Some features I spotted are:
        * Comm module is imported in mesh partition example
        * There's a parallel switch in numerical settings and it seems
        to be forcing flux and solver settings e.g. nLayersOfOverlapForParallel
        parallelPartitioningType, linearSolverConvergenceTest


At the end of the day I would like to run

mpirexec -np 10 python mesh_and_discretisation_test.py 

in order to accelerate calculations.

Please let me know,
Robert
-- 
Not everything that's true can be proven so!
http://en.wikipedia.org/wiki/Godel%27s_incompleteness_theorems

cekees:
Hi Robert,

Yes, no problem. I'm going to switch this  conversation over to the proteus
issue tracker and will CC you on that.

Thanks,
Chris

PR #935, opened 2016-03-08
Add the Adios library

certik:
https://www.olcf.ornl.gov/center-projects/adios/


certik:
(The failure is unrelated.)

PR #922, opened 2016-02-12
pytest: update to version 2.8.6 (pytest-cov needs >= 2.6.0)

johannesring:

PR #936, opened 2016-03-10
Bumping version of py as required by recent bump in py.test

mikaem:
See #922 
py is a dependence of py.test and recent bump requires py to be >= 1.4.29


johannesring:
Thanks.

PR #938, opened 2016-03-15
fftw2: supply the correct F77 compiler

certik:
Otherwise it uses the system wide one.


cekees:
Could you use $MPIF77 instead so gfortran isn't hardwired  in?

certik:
I need to test this. It looks like the configure script finds the correct mpi programs, but requires $F77 to be the actual compiler, not the MPI wrapper. So that's why it's tricky. I could use $FC, but sometimes $FC is not set...

certik:
The best solution is of course to make `hit` aware of compilers and set the `FC`, `F77`, `CC` and `CXX` variables properly. Then there is no problem.

Issue #395, opened 2014-07-29
Paraview fails with "warning: libbz2.so.1.0, needed by libfreetype.so.6, not found (try using -rpath or -rpath-link)"


certik:
One one cluster:

```
[paraview] Scanning dependencies of target vtkParaViewDocumentation
[paraview] [  8%] [  8%] Building CXX object Applications/ParaView/Documentation/CMakeFiles/vtkParaViewDocumentation.dir/ParaViewDocumentationInitializer.cxx.o
[paraview] Building CXX object Applications/ParaView/Documentation/CMakeFiles/vtkParaViewDocumentation.dir/qrc_paraview_documentation.cxx.o
[paraview] /usr/bin/ld: warning: libbz2.so.1.0, needed by /panfs/scratch1/certik/bld/freetype/3j5vs5xri633/lib/libfreetype.so.6, not found (try using -rpath or -rpath-link)
[paraview] /panfs/scratch1/certik/bld/freetype/3j5vs5xri633/lib/libfreetype.so.6: undefined reference to `BZ2_bzDecompress'
[paraview] /panfs/scratch1/certik/bld/freetype/3j5vs5xri633/lib/libfreetype.so.6: undefined reference to `BZ2_bzDecompressInit'
[paraview] /panfs/scratch1/certik/bld/freetype/3j5vs5xri633/lib/libfreetype.so.6: undefined reference to `BZ2_bzDecompressEnd'
[paraview] collect2: ld returned 1 exit status
[paraview] gmake[2]: *** [bin/TestingDemo] Error 1
[paraview] gmake[1]: *** [ThirdParty/QtTesting/vtkqttesting/Examples/CMakeFiles/TestingDemo.dir/all] Error 2
```

But

```
$ ldd /panfs/scratch1/certik/bld/freetype/3j5vs5xri633/lib/libfreetype.so.6
    linux-vdso.so.1 =>  (0x00007fffe974e000)
    libz.so.1 => /panfs/scratch1/certik/bld/freetype/3j5vs5xri633/lib/../../../zlib/3el5ccejre7b/lib/libz.so.1 (0x00002b85a1286000)
    libbz2.so.1.0 => /panfs/scratch1/certik/bld/freetype/3j5vs5xri633/lib/../../../bzip2/yhn7t7sdxdfd/lib/libbz2.so.1.0 (0x00002b85a149d000)
    libc.so.6 => /lib64/libc.so.6 (0x00002b85a16d2000)
[...]

PR #942, opened 2016-04-05
chrono: added bzip2 as explicit dependency

tridelat:
@cekees I got a similar error as https://github.com/hashdist/hashstack/issues/395 when trying to build chrono on CentOS.
Making bzip2 an explicit dependency for chrono fixed it.


cekees:
Thanks!

PR #943, opened 2016-04-13
Add libjpeg(-turbo)

vbraun:
Libjpeg is a requirement for pillow starting with version 3

There are multiple implementations, but libjpeg-turbo seems to be popular.

### DIFF
--- a/pkgs/host-jpeg.yaml
+++ b/pkgs/host-jpeg.yaml
@@ -1,0 +1,1 @@
+extends: [host_package]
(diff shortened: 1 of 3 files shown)


vbraun:
PS: The way I understand it, openjpeg only supports jpeg2000 so thats not good enough for pillow.

cekees:
+1. I was wondering what libjpeg  implementation to include.  Glad you
added  this.

PR #944, opened 2016-04-20
update numpy yaml to support version 1.11

smfcoding:
Somewhere between numpy version 1.8 and 1.11 the dependencies changed from distutils to cython
and setuptools.

So I am not sure how this type of dependency should be handled, because now the _extends_ section of  yaml configuration file has to be changed accordingly.


vbraun:
+1

PR #946, opened 2016-04-28
Proteus darwin 10.11.4 fixes

mfarthin:
Updates stack to compile with clang 7.3.0 on el Capitan. 
    - sets `proteus.Darwin.yaml` to extend `osx_clt.yaml`
    - removes all `fortran` dependencies in stack  
    - adds parmetis patch for clang compile issue (see pull request 3491 from 
      `homebrew/homebrew-science`)
    - defines `PETSC_SKIP_COMPLEX` to avoid template compile errors due to inclusion of `<complex>`
       in `petscmath.h`
    - sets `functools` to `host-functools` to avoid python build errors


mfarthin:
`jupyter` also seems to require `pip install pexpect gnureadline appnope`

certik:
This fails tests. Does it require a newer `hit`?

mfarthin:
Thanks, I will check with @cekees. It looks like there were similar failures with PR #942

cekees:
I merged this and fixed the travis tests. I switched the default stable/proteus stack to use hashstack mpich, though I usually use the host-mpich on ubuntu.

PR #947, opened 2016-04-29
patch MAX_JBUFS=24 in parmetis on OS X

minrk:
clang 7.3 requires MAX_JBUFS < 25


certik:
That looks good, thanks @minrk.

PR #954, opened 2016-05-16
Add missing libs to netcdf4 on Ubuntu 16.04

certik:


certik:
I think the Travis failure is just a slow Travis, since the `pr` and `push` tests should be equivalent.

certik:
I restarted it, now Travis passes.

PR #962, opened 2016-06-28
Update Scipy

certik:


vbraun:
I updated cython, if you merge/rebase master then it should work...

certik:
Thanks @vbraun for fixing it!

PR #964, opened 2016-07-06
Update various packages

certik:
* Disable installing Gtest in Trilinos
  This clashes with user's own installation of gtest.

* Upgrade gtest

* Make doxygen build

* Add unittest-cpp

* Use an alternative download for parmetis
  The original URL does not work.

* Exodues: Remove a patch (already applied)

* Exodus sources have moved


certik:
Test failures due to cython.org being down. I restarted the failed builds.

PR #965, opened 2016-07-17
Update to Cython 0.24.1

vbraun:
The previous download url does not work any more, and cython.org had a major service interruption a month ago. Use the pypi download url instead.


certik:
+1, thanks for fixing it!

PR #968, opened 2016-10-05
Update to openssl-1.0.2h and ruby-2.3.1

vbraun:
Straightforward updates


jcftang:
+1 assuming the tests pass this should be mergeable.

PR #971, opened 2016-11-02
stable/proteus: updated yaml files of pkgs for gcc 6 compilation

tridelat:
A couple of packages updates for compiling with gcc 6.2.1 (using arch linux).
When compiling with gcc 6, zoltan can to be commented out in `examples/proteus.linux2.yaml` as the current version throws an error


cekees:
Thanks!

PR #976, opened 2016-11-30
Add patch to build sharedlibs on cygwin

### LINKED ISSUE #367: The sw_vers -productVersion | sed "s/\(10.[0-9]\).*/\1/" line in osx.yaml will stop working for OS X 10.10
The line at https://github.com/hashdist/hashstack/blob/master/osx.yaml#L8:

```
    export MACOSX_DEPLOYMENT_TARGET=$(sw_vers -productVersion | sed "s/\(10.[0-9]\).*/\1/")
```

Will stop working once the OS X version is 10.10. So it needs to be improved.

Reported by @asmeurer.

embray:
This adapts the patch from http://trac.sagemath.org/ticket/17333 to allow libtool to make the shared lib in the first place rather than trying to hack it together after libtool is run.  While maintaining patches is always unfortunate this should be more reliable.  Fixes #367


embray:
Note, the issue is also fixed upstream via https://github.com/zeromq/libzmq/pull/1283 in libzmq v4.2.

cekees:
+1. Thanks!

cekees:
Did you run into any problems with the call to config.sub in the configure script? I tried some testing on cygwin and several autotools packages seem to be failing at that step.

embray:
I haven't yet run into any problems with `config.sub`, only `config.guess`.

Just as a heads up, I have a big patch for Python coming soon too.

PR #977, opened 2016-11-30
Fix incorrect example in the docstring for the configure build stage

embray:
The example had both incorrect syntax and does not match what is implemented in the code.

It might be nice the `configure` stage options were better documented too.


cekees:
Thanks, looks good to me. Are you guys testing on osx? Just curious as I had an issue with another package passing -rpath to the linker and generating an erro.

embray:
I am just personally trying to see how far I can get with hashstack on cygwin.

PR #978, opened 2016-11-30
Add patch for sqlite for config.guess to properly identify Cygwin-64 …

embray:
…when `uname -m` returns `'x84_64'`

Tried to do this without a patch--in princple this could be fixed by passing the correct --build flag
to configure, but it was too tricky to do that on a conditional basis.

This could also be fixed by giving sqlite a newer `config.guess` entirely but it's easy enough to add this one small patch.

### DIFF
--- a/pkgs/sqlite.yaml
+++ b/pkgs/sqlite.yaml
@@ -1,7 +1,0 @@
-extends: [autotools_package]
-dependencies:
-  build: [readline]
-
-sources:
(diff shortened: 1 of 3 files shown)


cekees:
+1. BTW, how are you building on cygwin? By hand or using some sort of CI? (Can you do cygwin builds in AppVeyor?)

embray:
Cygwin builds are possible on AppVeyor though I haven't tried it yet.  Once I get everything working locally I could help set that up if you want.

PR #982, opened 2016-12-07
Patch gdbm on Cygwin so that it can build a shared library

embray:
Title pretty much says it all.  This patch is quick and dirty since it's only applied on Cygwin, but without it gdbm does not build a shared library on Cygwin.

### DIFF
--- a/pkgs/gdbm.yaml
+++ b/pkgs/gdbm.yaml
@@ -1,8 +1,0 @@
-extends: [autotools_package]
-
-defaults:
-  relocatable: false
-
-sources:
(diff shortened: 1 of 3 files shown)


vbraun:
Whats the status of the cygwin PRs? I'd be happy to merge them if you say so...

embray:
I haven't really played with hashdist since then.  Should work AFAIK.

cekees:
Do you recall which profile you use to build on cygwin?

embray:
Just the default one IIRC: https://github.com/hashdist/hashstack/blob/master/cygwin.yaml

PR #985, opened 2017-02-07
Build Python 2.7 with UCS-4 enabled

vbraun:
Without it, Python cannot handle regular expressions involving 4-byte
unicode chars. This is also the default in Python >= 3.3

I just tripped over this:
~~~~
In [1]: import re
In [2]: re.compile(u'[\U00010000-\U0010ffff]')
error: bad character range
~~~~
Whereas it works with the system python2 and python3 on Fedora, probably all other distros too.

### DIFF
--- a/pkgs/python/python.yaml
+++ b/pkgs/python/python.yaml
@@ -73,6 +73,11 @@ build_stages:
     mode: update
     extra: ['--without-ensurepip']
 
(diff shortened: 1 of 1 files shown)


vbraun:
I think its pretty uncontroversial, merging...

PR #987, opened 2017-03-30
Change zeromq download url to github

vbraun:
The old zeromq download url is apparently defunct. While we are at it,
update to the latest legacy stable release.


vbraun:
Should be uncontroversial...

PR #990, opened 2017-05-25
updated chrono and gmsh pkgs source

tridelat:
Chrono commit 19c1e55: works with proteus branch in development using chrono functions
Gmsh version 3.0.2: to pass gmsh related tests in proteus


tridelat:
@cekees , I am not sure why the Travis build fails here, could this PR still be merged as it is or is there something that should be modified?

PR #991, opened 2017-06-19
Add Mac fixes to hashstack's proteus configuration.

tjcorona:


pkgs/doxygen/doxygen.yaml
```diff
@@ -7,11 +7,11 @@ sources:
...
 
-build_stages:
-  - when: platform == 'Darwin'
-    name: patch
-    before: configure
-    files: [libiconv_plug.patch]
-    handler: bash
-    bash: |
-      patch -up1 < _hashdist/libiconv_plug.patch
+#build_stages:
```
cekees: you can just delete this commented code if it's not necessary any more

pkgs/h5py.yaml
```diff
@@ -14,6 +14,7 @@ build_stages:
   before: install
   handler: bash
   bash: |
-       export HDF5_VERSION='1.8.16'
+       export HDF5_VERSION='1.8.13'
```
cekees, line 17: What necessitated going backward on the release?

pkgs/scorec.yaml
```diff
@@ -3,9 +3,9 @@ dependencies:
   build: [mpi, parmetis, zoltan]
 
 sources:
-  - url: https://github.com/SCOREC/core.git
-    key: git:e219777edd875d35096dfedd7197300b8e2f3837
-    
+  - url: https://github.com/tjcorona/core.git
```
cekees, line 6: Have your changes to scorec core not been merged in? Maybe we should put and if statement here to do this conditionally only for darwin or do it as a patch instead. Maybe @zhanga can look into the changes on scorec core.

cekees (approved):
I had a few questions above, but I think we can get this merged in soon and get the stack commit updated on proteus master.

pkgs/h5py.yaml
tjcorona: This change was made by @mfarthin.

pkgs/scorec.yaml
tjcorona: I think the version we were using had the error, and the newest version of SCOREC/core wasn't compatible with other things. The easiest fix was to branch and patch the version we were using.

PR #999, opened 2018-04-18
Update SCOREC pkg for 2D Reconstruction

zhang-alvin:
This update is to get erdc/proteus#708 building properly.


zhang-alvin:
@cekees Another update to the scorec.yaml for erdc/proteus#708

cekees:
I switched from the git to zip protocol because of issues we've been seeing on HPC machines related to SSL. It's more efficient anyway to use a single source archive. The git approach in hashdist is less efficient and the archives can't be stored on the remote cache, so you should only use it while developing--switch to an archive format before pushing to the stack.

For future reference, one way you do this is to run `hit fetch` on the URL for the zip archive of the commit, which you can get by constructing the URL from the formula or copying the URL of the "download zip" link from browsing the commit source: https://github.com/SCOREC/core/tree/fc94c3d4b013d7535e43549577b2c5f619d43d4b

zhang-alvin:
Okay, that sounds fine.

